PR #5, opened 2015-09-24
Add Nikon 1 S2

thusoy:
I'm not sure if I'm including all the details you're after, but this is what I did:
  - Identified the usb id of the camera with `lsusb`
  - Added the id to `ptp2/library.c` with the same capabilities as the previous model (S1)
  - Compiled
  - Ran `--summary` and `--list-all-config` and added the results as cameras/nikon-s2.txt
  - Removed the serial number of the camera

Scream and shout if you need me to add anything else.

### DIFF
--- a/camlibs/ptp2/cameras/nikon-s2.txt
+++ b/camlibs/ptp2/cameras/nikon-s2.txt
@@ -1,0 +1,402 @@
+Camera summary:
+Manufacturer: Nikon Corporation
+Model: S2
+  Version: Ver.1.0100
+  Serial Number: 000000000000000000000000nnnnnnnn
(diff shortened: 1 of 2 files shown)


msmeissn:
This was very good, thanks!

one more thing, can you extract the debug log, 

gphoto2 --summary --debug --debug-logfile=xx.log

and the device info part between
 >>>>>>>>>>>>>>
"Device info:"
"Manufacturer: Nikon Corporation"
... and
"Device Properties Supported:"
... hex ids ... 0x5xxx and 0xdxxx
<<<<<<<<<<

so we know the exact opcode and propertylist

thusoy:
Certainly, added a new PR with the opcodes.

PR #9, opened 2015-10-03
Update German translation

nazgul77:


msmeissn:
you just converted from latin1 to utf8 right? if so, please change the logmessage accordingly and also please adjust the PO Header

"Content-Type: text/plain; charset=ISO-8859-1\n"

nazgul77:
No, it is actually just a one-line change because otherwise gtxml check fails on the PO file.
I did not intend to change the encoding. Sounds like the character encoding currently is different from what is stated in the header. Let me look into this.

PR #35, opened 2016-01-14
Avoid overlinking of libexif

whydoubt:
The dependency on libexif need not be exposed to consumers of this
library.  This is fixed by moving libexif from Requires to
Requires.private in the pkg-config files.


msmeissn:
hmm. there must have been a reason, but lets see what breaks

PR #47, opened 2016-03-24
handle negative exposurecompensation settings

carycode:
* connect a USB cable to a Sony alpha 7s
* At the command line, run

    gphoto2 --set-config exposurecompensation=-1

* What I expected to happen:

I expected the exposure compensation value displayed on the
status line at the bottom of the little screen on the back of the camera
to change to "-1.0".

* What actually happens:

The exposure exposure compensation value displayed on the
status line at the bottom of the little screen on the back of the camera
increments up all the way to the maximum.
(In "movie" mode, it stops at "+2.0").

This patch seems to fix things so that command line

    gphoto2 --set-config exposurecompensation=-1

does what I expect.

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -1476,7 +1476,7 @@ _put_sony_value_u16 (PTPParams*params, uint16_t prop, uint16_t original_desired_
                 }
         } while (1);
         return GP_OK;
-
+}
 
 
 static int
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -1383,8 +1383,101 @@ _put_sony_value_u##bits (PTPParams*params, uint16_t prop, uint##bits##_t value,i
 	return GP_OK;									\
 }
 
+#if 0
 PUT_SONY_VALUE_(16) /* _put_sony_value_u16 */
+#endif
 PUT_SONY_VALUE_(32) /* _put_sony_value_u32 */
+
+/*
+2016-03-23: David Cary at Lamar Systems:
+tweak to get positive and negative exposure compensation values working.
+Replaces "PUT_SONY_VALUE_(16)"
+*/
+static int
+_put_sony_value_u16 (PTPParams*params, uint16_t prop, uint16_t original_desired_value, int useenumorder) {
+        GPContext               *context = ((PTPData *) params->data)->context;
+        PTPDevicePropDesc       dpd;
+        PTPPropertyValue        propval;
+        uint32_t                origval;
+        uint32_t                value = (int32_t)(int16_t)original_desired_value;
+        time_t                  start,end;
+
+        GP_LOG_D("setting 0x%04x to 0x%08x", prop, value);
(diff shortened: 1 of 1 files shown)


msmeissn:
it does nort build ... but its probably just a misisng }

carycode:
Yes, missing a "}". Fixed.

The effect of this patch is to always treat the value passed to _put_sony_value_u16() as a *signed* 16 bit number (rather than the previous *unsigned*).
That fixes the "gphoto2 --set-config exposurecompensation=-1" bug.
(Any tips on how to test to make sure this doesn't break something else?)

The "`PUT_SONY_VALUE_(bits)`" macro was used twice by gphoto2, to generate the "`_put_sony_value_u16()`" and "`_put_sony_value_u32()`" functions.

To make it easier to debug the problem, I temporarily made `_put_sony_value_u16()` a normal function.
Is it better to leave it as a separate function, or to refactor and merge `_put_sony_value_u16()` back into the "`PUT_SONY_VALUE_(bits)`" macro?

msmeissn:
it should be refactored and merged, but I see that I can do that next week

msmeissn:
i did a different change, adding a third i16 method.

it also goes below 0 for exposure compensation now.

however on my alpha 51 when I try to set "2.3" it flips between 2.0 and 2.3 and does not terminate.

same for 1.3, -1.3 and -2.3, -0.7, 0.7.

this is something left to fix

carycode:
Awesome. Thank you.
I can pick up this Sony alpha 7s and (by hand) use the menu buttons to set exposure compensation "half-step mode" (... -1.5, -1.0, -0.5, 0, 0.5, ... etc.) or "third-step mode" (... -0.7, -0.3, 0, 0.3, 0.7, 1, ... etc.). No matter which state the camera is in, when gphoto2 asks the camera what settings are available ("gphoto2 --get-config exposurecompensation"), the camera gives the same list that includes both half-steps and third-steps. When I try to set ("gphoto2 --set-config exposurecompensation") a half-step value when it's in "third-step" mode, it flips back and forth between the nearest third-step values forever. When I try to set a third-step value when it's in half-step mode, it flips back and forth between the nearest half-step values forever. Is that the same glitch you see? (Should I post this as a separate bug report?) I wish the camera supported some command so gphoto2 could tell the camera to go into half-step mode, or the camera supported a gphoto2 command to go to a "half-step" position even if the camera was in "third-step position". Then I wouldn't have to remember to manually set "half-step mode" first every time I want to plug the USB cable into the camera.

msmeissn:
i made the selection routine more resistant to float rounding errors. perhaps this will help yo utoo..

if not, can you open a new issue ... my a58 does not have half step modes though :/

carycode:
Thank you. The latest gphoto2+libgphoto2 I downloaded now sets negative exposurecompensation settings properly on this Sony alpha 7s. I'll test half-steps and third-steps later and open a new issue if I notice float rounding errors. This issue has been fixed and can be closed. Thank you again.

PR #64, opened 2016-06-15
jpeg_memsrcdest: extend feature check

pohly:
libjpeg.h in OpenEmbedded master (from libjpeg-turbo 1.5.0) provides
these methods if "JPEG_LIB_VERSION >= 80 ||
defined(MEM_SRCDST_SUPPORTED)".

The support for the jpeg_mem functions was added even when not
emulating the libjpeg8 API, controlled via the MEM_SRCDST_SUPPORTED
define, so checking for the version alone is not enough anymore.

See https://github.com/libjpeg-turbo/libjpeg-turbo/commit/ab70623eb29e09e67222be5b9e1ea320fe5aa0e9

This fixes errors about conflicting declarations (signed vs. unsigned
char).

Signed-off-by: Patrick Ohly <patrick.ohly@intel.com>

### DIFF
--- a/camlibs/ax203/jpeg_memsrcdest.c
+++ b/camlibs/ax203/jpeg_memsrcdest.c
@@ -25,8 +25,8 @@
 #include "jpeg_memsrcdest.h"
 
 /* libjpeg8 and later come with their own (API compatible) memory source
-   and dest */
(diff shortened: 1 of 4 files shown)


msmeissn:
thanks! :)

PR #69, opened 2016-07-17
Adding capabilities for Nikon COOLPIX A

philip222:
Added commands for Nikon COOLPIX A which the camera doesn't list, but I've confirmed it does in fact support.  Allows for use of this camera, otherwise this camera doesn't work.


msmeissn:
thanks

PR #76, opened 2016-08-15
Fixes for Canon 80D and 750D

xaphod:
Fix for Canon 80D: when adding objects, Canon 80D uses different event-code
Fix for Canon 750D: during live view, Canon 750D uses different blob code for JPG data

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -2561,6 +2561,7 @@ camera_capture_preview (Camera *camera, CameraFile *file, GPContext *context)
 						continue;
 					case 9:
 					case 1:
+                    case 11:
 						if (len > (size-(xdata-data))) {
 							len = size;
 							GP_LOG_E ("len=%d larger than rest size %ld", len, (size-(xdata-data)));
--- a/camlibs/ptp2/ptp-pack.c
+++ b/camlibs/ptp2/ptp-pack.c
@@ -1828,6 +1828,7 @@ ptp_unpack_CANON_changes (PTPParams *params, unsigned char* data, int datasize, 
 		ce[i].u.info = NULL;
 		switch (type) {
 		case  PTP_EC_CANON_EOS_ObjectAddedEx:
+        case  PTP_EC_CANON_EOS_ObjectAddedUnknown:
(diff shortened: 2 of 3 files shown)


kentavv:
Hi Tim, when I saw your patch I quickly gave it a try (unfortunately the results didn't change.) Are these the only changes that you expect are required for libgphoto2 to work well with the Canon 80D? Thank you

msmeissn:
thanks! (could have sent two pull request for the different things, but its ok.)

msmeissn:
this patch will not cure a "just works once after connect" problem.

xaphod:
@kentavv what are you trying to do that is failing? If liveview I can think of 2 things gphoto2 probably isn't doing right for the 80D:

1. When starting liveview with PTP_OC_CANON_EOS_GetViewFinderData, gphoto2 is using just 1 param of 0x00100000. The Canon app (and my project) use 3 params like this: ptp_init_container(&ptp, PTP_OC_CANON_EOS_GetViewFinderData, 3, 0x00200000, 0x00000001, 0x00000000);

2. Taking photos in liveview: instead of sending a shutter release (PTP_OC_CANON_EOS_RemoteRelease), you need to instead send PTP_OC_CANON_EOS_RemoteReleaseOn then PTP_OC_CANON_EOS_RemoteReleaseOff (different codes). Perhaps the 80D requires this, I don't think gphoto2 is doing this

xaphod:
@msmeissn you are very welcome, thanks for this awesome project! Sorry I just noticed my tabs were off -- will fix that if I check in again...

msmeissn:
2. we do not remotereleasedon/remotereleaseoff if available these days.

1. ... it seems to work with just one parameter, have you decoded what the paramerters mean to PTP_OC_CANON_EOS_GetViewFinderData,

kentavv:
@xaphod Two commands that I've tried are --summary and --capture-image-and-download. The --summary command will return information once after attaching the camera, but each time after there is a PTP error. The --capture-image-and-download command will activate the shutter once after attaching the camera, but not again without re-attaching. On the first call, the image is never downloaded and gphoto2 waits. Subsequent calls return a PTP error. Looks like @msmeissn knows the problem. I'm impressed both of you have such in depth knowledge of the camera protocols. That has to represent 1000s of hours of work.

Issue #87, opened 2016-09-13
gp_camera_capture() taking more than one photo Sony ILCE-5000


lluk12:
Is it normal that when setting the Sony ILCE-5000 in Spd. Priority Cont or Cont. Shooting it takes more than one photo?

Instead of taking one photo and continuing the execution, it takes two really fast, so when I want to get another picture, I get the second photo instead of taking a new shot.

The problem is that, in Single Shooting mode I get 2second delay between images, but in the other modes, I get 2 images with less than 0'5 seconds between them, and then 1 second gap until the next two photos.

Thanks!

msmeissn:
can you test current git, as the strategy for sony capture was changed?

lluk12:
Yes, just installed the new version. I'll give you feedback as soon as I can try it (I don't have the camera right now).

lluk12:
It does work now. When I call gp_camera_capture() it just makes one photo. Apart from that, it's much more easy to control the times now.

Thanks for the quick fix.

msmeissn:
the summary has "Sony Capture" listed ...are you expecting something different ? 

thanks to @elijahparker for the sony improvements :)

lluk12:
Note that "Speed Priority Cont" mode still has the same issue, even increasing the usleep() time to 2 seconds in line 3675

PR #88, opened 2016-09-13
Fixed sony capture reliability issues. Tested with Sony A7SII

elijahparker:
Rather than waiting for the new object event (which often is simply not emitted by the camera), the previously undocumented property 0xD215 is used.  When no image is in the camera's RAM, 0xD215 has a u16 value of 0x0. However, when an image is in RAM for download, the u16 value is 0x8001.  If two images are available, it's 0x8002, etc.  After the images are retrieved (downloaded), the 0xD215 property returns to 0x0.

The sony_capture method has been updated to use this new method instead of the unreliable new object event.  The timing for the capture commands has also been modified to reflect how the Sony RemoteCameraControl app initiates it.  This should also fix issue #87 .

Some background on the fix is provided here: https://www.kickstarter.com/projects/elijahparker/view-intervalometer-auto-ramping-and-wifi-camera-r/posts/1677561

This has been tested with the Sony A7SII only.  I certainly hope this fix applies to the other models, but nothing can be certain.  More testing should be done if possible.

Thanks for the great software!

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -3651,9 +3651,6 @@ camera_sony_capture (Camera *camera, CameraCaptureType type, CameraFilePath *pat
 	PTPDevicePropDesc	dpd;
 	struct timeval	event_start;
 
-	propval.u16 = 1;
-	C_PTP (ptp_sony_setdevicecontrolvalueb (params, PTP_DPC_SONY_AutoFocus, &propval, PTP_DTC_UINT16));
-
 	C_PTP (ptp_generic_getdevicepropdesc (params, PTP_DPC_CompressionSetting, &dpd));
 
 	GP_LOG_D ("dpd.CurrentValue.u8 = %x", dpd.CurrentValue.u8);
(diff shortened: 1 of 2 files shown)


msmeissn:
The patch looks very good, ... applied, thanks!

msmeissn:
I had to do some incremental fixes.

the sony a58 did not take images anymore, the 50ms wait was too short.  

I debugged a bit and it seems the 0xd213 property indicates focus or capturing when it changes from 1 to 2.  I have added a check for that in the press logic.
I also added back event checking additionally on top of the property checking, and also read the properties.

elijahparker:
Nice work with the 0xd213 property -- that seems like a very robust way to handle the camera-specific timing requirements rather than a hard-coded delay.

Regarding the event checking, it might just be my system, but I first tried to use both the event checking and watching the 0xd215 property but I still had occasional issues when the event was trigger just prior to the property change.  I noticed the Sony app would always wait for the 0xd215 property even if the event fired first.

I at least know the current code won't run reliably on my system with the Sony A7SII.

Again, it might be that my embedded system just wasn't handling the events well, but the best solution might be to check is the 0xd215 property exists and if so, don't use events. (Or does it even exist on the a58 but not do anything?)

Thanks again for all your hard work on this project.

msmeissn:
d215 exists and shows the correct behaviour on the a58, so we can rely on that for now.

interesting to see the events gets there early. 

I have to cehck if that fixes my camera spontaneous reboots after capture :/

msmeissn:
camera is more stable when i do not use the event based method and just use the d215 method... :)

elijahparker:
Great to hear!  That aligns with what I observed during testing with the A7SII.

PR #100, opened 2016-10-15
Windows port using MSYS2 MINGW64

peterbud:
This is a basic port of the library to Windows. I have used the MSYS2 environment to compile the source. I have made only minimal changes necessary, please check whether the *nix builds are still working after the changes. The gphoto setting file should be placed in Windows under the current users profile folder, the existing code tried to save the settings file to the WIndows System directory which is was not working, as system prohibits that since a few years. Along the way I have fixed some undeclared variable errors.

The following functions are working:
- **gp_camera_init** and **gp_camera_get_summary** and **get_config_value_string**, tested by sample-autodetect.exe
- **gp_camera_capture**, tested by sample-capture.exe

The sample-trigger-capture.exe does not work yet, I'll open a question on that one separately.
I'm also going to port the gphoto2 as well, just need more testing.

I have used my Nikon D5000 to test, it is really fun working with this libary.

Peter

### DIFF
--- a/examples/samples.h
+++ b/examples/samples.h
@@ -13,4 +13,9 @@ int canon_enable_capture (Camera *camera, int onoff, GPContext *context);
 extern int camera_auto_focus (Camera *list, GPContext *context, int onoff);
 extern int camera_eosviewfinder (Camera *list, GPContext *context, int onoff);
 extern int camera_manual_focus (Camera *list, int tgt, GPContext *context);
+
+#if !defined (O_BINARY)
+    //To have portable open() on *nix and on Windows
+    #define O_BINARY 0
+#endif
 #endif
--- a/.gitignore
+++ b/.gitignore
@@ -119,3 +119,4 @@ stamp-h1
 tests/ddb/check-ddb.sh
 tests/test-filesys
 tests/test-gphoto2
+.vscode/
--- a/camlibs/pentax/pslr.c
+++ b/camlibs/pentax/pslr.c
@@ -1021,7 +1021,7 @@ static int ipslr_press_shutter(ipslr_handle_t *p)
     CHECK(ipslr_write_args(p, 1, 2));
     CHECK(command(p, 0x10, 0x05, 0x04));
     get_status(p);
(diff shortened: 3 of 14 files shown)


msmeissn:
this is too much in one. I need to untangle it manually, or if you want to submit it in parts this wuld help.

like e.g. the examples/ part seperate and others , perhaps per directory. 


i fixed pslr.c and sonydscf1 already and will slow go through the rest too

examples/samples.h
```diff
@@ -13,4 +13,9 @@ int canon_enable_capture (Camera *camera, int onoff, GPContext *context);
 extern int camera_auto_focus (Camera *list, GPContext *context, int onoff);
 extern int camera_eosviewfinder (Camera *list, GPContext *context, int onoff);
 extern int camera_manual_focus (Camera *list, int tgt, GPContext *context);
+
+#if !defined (O_BINARY)
```
msmeissn: is O_BINARY really needed to save files in binary mode? or would this just be for fopen?

peterbud:
And apologies I did not mean to overwhelm you ;) Take your time. 
I have managed to add the library to the official MSYS2 packages using the same patches:
[https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-libgphoto2](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-libgphoto2)

msmeissn:
I merged most of the things now, only  libgphoto2_port/libgphoto2_port/gphoto2-port-portability.c
i have not merged so far, it is a bit weird.

peterbud:
Great news Markus, thank you.
For the gphoto2-port-portability.c, as you saw my comment, it was also unclear to me what was the purpose of this code block: 
```
 if (path[0] != '.') {
                path[0] = path[1];
                path[1] = ':';
                path[2] = '\\';
        }
```
if the path starts with a '.' why it is needed to copy the second character to the first place? 

As I see this was only compiled if there was a WIN32 build, so I guess it has never been tested properly. For me this has been resulted always erroneous file names.

What would be your suggestion, how to handle this?

PR #113, opened 2016-12-05
Cleanup of camera_summary()

axxel:
A set of 4 patches implementing some code cleanup/cosmetic related to the `camera_summary()` function in ptp2/library.c. It is mainly about reducing code duplication and noise (removing 100+ lines of code), improving maintainability. Tested with a EOS 5Dm4 -> outputs the exact same (English) text as before.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -5170,225 +5170,161 @@ static int
 camera_summary (Camera* camera, CameraText* summary, GPContext *context)
 {
 	unsigned int i, j;
-	int n;
 	uint16_t ret;
-	int spaceleft;
-	char *txt;
+	char *txt, *txt_marker;
 	PTPParams *params = &(camera->pl->params);
 	PTPDeviceInfo pdi;
 	PTPStorageIDs storageids;
 
 	SET_CONTEXT(camera, context);
 
-	spaceleft = sizeof(summary->text);
 	txt = summary->text;
 
-	n = snprintf (txt, spaceleft,_("Manufacturer: %s\n"),params->deviceinfo.Manufacturer);
-	if (n >= spaceleft) return GP_OK; spaceleft -= n; txt += n;
-	n = snprintf (txt, spaceleft,_("Model: %s\n"),params->deviceinfo.Model);
-	if (n >= spaceleft) return GP_OK; spaceleft -= n; txt += n;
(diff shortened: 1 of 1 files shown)


msmeissn:
thanks! was a bit on my wishlist for rewrite already.

msmeissn:
axel ... its not working as expected.

snprintf returns the amount of bytes that would have been written, not 0 if the size is exceeded.

axxel:
Ouch, that is a bit weird though, because I specifically tested this by artificially reducing the size of `summary->text` to make it too small for my standard EOS test case and it worked (that was a while back, though, this whole patchset was still lying around and I thought better publish it than throw it away.)

My suggestion would be to replace the `SPACE_LEFT` macro along the lines of:
`#define SPACE_LEFT max(0, summary->text + sizeof (summary->text) - txt)`

(Disclamer: can not access my compiler right now, so this is untested.)

This would effectively make sure the whole thing does what I thought it does: don't append any more characters as soon as `summary->text` is full. Repeatedly calling `snprintf` and incrementing `txt` will not make any difference to the result of `camera_summary`.

axxel:
I did a little test to recheck what I claimed above. It behaves just as I thought (clang and gcc under Linux), even though the `snprintf` would suggest the '3' should be overridden with the second `APPEND_TXT` call it is not the case. Adding a "MAX(0, ...)" to the `SPACE_LEFT` simply replaces the '-2' with a '0', the resulting string is the same.

```c
#include <stdio.h>

typedef struct {
	char text [3];
} CameraText;

int main()
{
	CameraText s = { "123" };
	CameraText* summary = &s;
	char* txt = summary->text;

// -1 so we can check if '3' in our string gets overridden
#define SPACE_LEFT (summary->text + sizeof (summary->text) - txt - 1)
#define APPEND_TXT( ... ) txt += snprintf (txt, SPACE_LEFT, __VA_ARGS__)

	printf("%ld\n", SPACE_LEFT); // output = 2
	APPEND_TXT( "ab" );
	printf("%ld\n", SPACE_LEFT); // output = 0
	APPEND_TXT( "cd" );
	printf("%ld\n", SPACE_LEFT); // output = -2

	printf("%s\n", summary->text); // output = "a"
	printf("%c\n", s.text[2]);  // output = "3"

	return 0;
}
```

PR #128, opened 2017-01-10
Sony liveview support, sony capture reliability

elijahparker:
I've found out that liveview is indeed possible with many newer Sony Alpha cameras.  It simply requires fetching an object (jpeg) from 0xffffc002 (for comparison, captured images are fetched from 0xffffc001).

A challenge I ran into when implementing this was that objectinfo does not return anything useful for 0xffffc002, so I created a copy of ptp_getobject to include the size of the object fetched so that it can be handled correctly (ptp_getobject_with_size).  This seems to work well, however, feel free to take a different approach if something else will better fit the existing APIs.

Also, I have commented out the code in sony_capture for checking for the focus event -- in my testing, any use of events with the newest generation of Sony Alpha cameras causes it to occasionally hang (e.g., the Sony A7S seems ok, but the Sony A7SII hangs occasionally).  Based on the comments in the code, it seems this was necessary for older ones like the A58.  We might need to create a special case based on the presence of certain properties or something to handle this correctly. (Old camera case with events vs. new camera case without events).  For now, feel free to remove the #if 0..#endif found in this PR.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -2774,6 +2774,53 @@ enable_liveview:
 		SET_CONTEXT_P(params, NULL);
 		return GP_OK;
 	}
+	case PTP_VENDOR_SONY: {
+		uint32_t	preview_object = 0xffffc002; /* this is where the liveview image is accessed */
+
+		unsigned char *ximage = NULL;
+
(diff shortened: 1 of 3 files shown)


msmeissn:
thanks! I will see if it works or If I can adapt some smaller things

PR #144, opened 2017-03-16
Change PTP waif_for_event() timeout management for standard cameras

Kev-J:
Hi,

I just wanted to share some modifications I've done in the timeout management in PTP wait_for_event function : 
 7c95ede - to avoid error printed when the gp_port_check_int() return a timeout during check_fast. I suppose it isn't an error since the check_fast purpose is a non-blocking check on interrupt endpoint.
 ae93b58 - I checked the timeout value if the function fallback in "sonyout" (others than Canon and Nikon ?) as it is done for Canon and Nikon cameras.

Do not hesitate to comment.

Cheers,


msmeissn:
it would be better to set the timeout and wait for the event instead of looping.

but this is also tricky as the vendor event checks for nikon also should be run regulary, so this longer wait can only be used if there are no vendor  specific  event checks.

I can also do that logic if I find a free slot of time.

FWIW, camera_wait_for_event can return before the timeout expires and your code needs to handle that.

PR #171, opened 2017-06-02
fix memory corruption in gp_log_remove_func

axxel:
The gp_log_remove_func implementation had 2 severe issues:
* it moved way to few bytes
* it moved the wrong bytes to the wrong place, destroying libc memory
management structures (resulting in different types of crashes).

When the first item has to be removed, it moved a couple bytes from the
start of the array to the left (before the start of the array), instead
of moving the second and following items over the first one.

DISCLAMER: I had to quickly workaround the issues by disabling my gp_log_remove_func usage. This one-liner is untested. Please double check.


msmeissn:
I added trivial testcase in test-filesys.c, it no longer crashes. :)

PR #188, opened 2017-07-21
Added USB IDs: Canon SL2, Ricoh Theta S/SC

lacyrhoades:
Added USB ID for Canon EOS SL2 / Kiss X8 / 200D
Added USB ID for Ricoh Theta S and SC models
Added Git to install/build instructions


msmeissn:
thanks!

i removed the SVN reference completely, as its not used anymore.

PR #189, opened 2017-07-25
Improve error reporting when using the ptp2 lib 

fviard:
My goal with this PR is to improve the error reporting when using ptp2 (usb).
A patch to gphotofs will follow that take usage of this improvement.

There are 3 points of this PR:
- raise usb errors that are currently masked with a generic "PTP_ERROR_IO".
- add a new PTP_ERROR_NODEVICE to be able to pass through the missing device error detected by libusb.
- reworked event manager async queue for multiple purpose. (Main one being to not hide errors)

Please see the description message of each commit for more details.


msmeissn:
thanks!

lets hope nothing breaks :)

Issue #198, opened 2017-08-23
Canon FlexiZone AF support


evertheylen:
I'm using a Canon 1300D, which has a feature called FlexiZone Autofocus (like many Canons I suspect). Basically, it's a rectangle on your screen that you can move around with the arrows (comparable to touch to focus on touch screen devices). Here's an example:

![img_20170823_173543](https://user-images.githubusercontent.com/2071542/29624556-aa92f776-8829-11e7-8b22-655f9ef923ce.jpg)

I'm wondering whether I can control this through libgphoto. I know how to trigger the autofocus, but not how to move the rectangle. I've tried `gphoto --list-all-config` before and after moving the rectangle and diff'ed the output, without success.

At this point, I'm thinking it's probably not exposed through USB. However, could it be that there is some vendor-specific extension to PTP that would allow this? With some guidance, I'd be very happy to implement this myself.

For reference, I added the output of `--list-all-config`: [canon1300D_config.txt](https://github.com/gphoto/libgphoto2/files/1246034/canon1300D_config.txt)

Great library btw ;)

msmeissn:
can you capture debuglog before and after.

we might just not be decoding the values yet.

evertheylen:
Thanks for your response! Here are the logs, captured with `gphoto2 --debug --debug-logfile=square-location-debug.txt --debug-loglevel=all --list-all-config`:

  1. [square-topleft-debug.txt](https://github.com/gphoto/libgphoto2/files/1248097/square-topleft-debug.txt)
  2. [square-mid-debug.txt](https://github.com/gphoto/libgphoto2/files/1248098/square-mid-debug.txt)
  3. [square-bottomright-debug.txt](https://github.com/gphoto/libgphoto2/files/1248101/square-bottomright-debug.txt)

evertheylen:
I tried to do the same diff trick with the debug logs (after stripping away the timestamps), but I didn't find anything that could resemble the flexizone rectangle location. Is there another layer of code where the values/settings might be discarded, or do the debug logs already show every possible setting available to us?

msmeissn:
i also diffed it, no diff excepting the d17c aka EOS_UTCTime.

either we need to request it seperately, or there might be commands handling it.

can you try to gfet a trace of this on windows?

evertheylen:
I do have a Windows computer available, but I'm unsure of what you want. Is it related to the windows SDK Canon provides, or do I just run the same command as before but on windows?

[...]

PR #199, opened 2017-08-24
White balance adjust on Canon 1300D

evertheylen:
On my Canon 1300D, I was unable to adjust the white balance to negative values (both `whitebalanceadjusta` and `whitebalanceadjustb`). This is problematic, since it would limit the possible WB settings to a quarter of the possibilities. The horizontal Blue-Amber axis is controlled by `whitebalanceadjusta` (positive values are amber), the vertical Green-Magenta axis is controlled by `whitebalanceadjustb` (positive values are green).
  
White Balance Shift

When trying to set negative values, I got a "Device is busy" error (code -110). Getting the value worked fine (if set on the camera UI).

After some debugging I found out that it did work when I used a 32 bit integer instead of a 16 bit one. I tried to properly rewrite this to the best of my abilities, but I could not get the `_get` function to 'automatically' detect the possible values. I have no idea where they are set and how I could modify this in a clean way, so any feedback is welcome :) . Currently, I just hardcoded it.

Also, I have no idea what implications this may bring for other Canons. More testing might be needed, but I only have a Canon 1300D available.

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -4019,10 +4019,11 @@ _get_Canon_EOS_WBAdjust(CONFIG_GET_ARGS) {
 	gp_widget_new (GP_WIDGET_RADIO, _(menu->label), widget);
 	gp_widget_set_name (*widget, menu->name);
 
-	for (i = 0; i<dpd->FORM.Enum.NumberOfValues; i++) {
-		sprintf (buf, "%d", dpd->FORM.Enum.SupportedValue[i].i32);
+	for (i = 0; i<19 /*dpd->FORM.Enum.NumberOfValues*/; i++) {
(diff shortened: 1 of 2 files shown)


msmeissn:
can you attach a debuglogfile for --list-all-config ?

evertheylen:
They won't differ much from the ones in my issue: #198. Those are using the latest release, here is a debug log using my branch, with the setting B8, M4.
[B8-M4-debug.txt](https://github.com/gphoto/libgphoto2/files/1251423/B8-M4-debug.txt)

msmeissn:
hmm. your 1300D sends 32bit entities for those values, while the older ones send 16bit entities. I need to think how to handle this commonly for both

msmeissn:
i will revert the last patch as I fixed that issue, rest is ok

PR #200, opened 2017-08-28
More fixes for support 64bit getpartialobject for Android in read_file_func

fviard:
Fixes the assert that was in "C_PARAM_MSG" to be correct.
Added more checks.
Return GP_ERROR_NOT_SUPPORTED only if the "partialread" feature is not supported and so fallback should kick in upper level. Otherwise raise errors.

IMPORTANT NOTE: Currently this is not enough to read large files (>64 bits).
Because we are looking at the "compressedSize" that is limited to 32 bits.


msmeissn:
ok. will clean the whitespace change in a later commit

PR #207, opened 2017-10-16
prevent redefining CAMLIBS and IOLIBS on WIN32

vividos:
This change enables WIN32 users (MinGW and also Visual Studio, for which I may add some more pull requests soon) to redefine CAMLIBS and IOLIBS in the config.h file. This is useful in order to place the camlibs and iolibs shared libraries in subfolders, so that the file system is more cleaned up. Comments welcome!

### DIFF
--- a/gphoto2/gphoto2.h
+++ b/gphoto2/gphoto2.h
@@ -38,8 +38,9 @@ extern "C" {
 #endif
 
 #ifdef WIN32
-#undef CAMLIBS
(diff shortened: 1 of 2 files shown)


vividos:
@msmeissn Could you take a look at this patch? Does it match the expectation that I have of the two macros? Thanks!

msmeissn:
it changes behaviour a bit, but ok.

PR #261, opened 2018-04-11
Canon Object Change Support

littlexiang:
I've test several model and find that for camera file change case like "Rating" or "Protect"
under firmware <=1.1.0 the event is PTP_EC_CANON_EOS_ObjectInfoChangedEx (tested with D1300, ver 1.1.0, purchased recently,  I think it's the most cheap model now)
and above firmware >=1.1.3 the only event is PTP_EC_CANON_EOS_ObjectContentChanged (tested with 1DX Mark II, ver 1.1.3 & 1.1.6 purchased mid 2017)

so this PR is an upgrade to my last PR, 
it's now compatible for both events and returns the same result of object path and filename


msmeissn:
itwould be cool if you could rebase uyour branch against master and send another pull request. you can also merge the fix double free into the original commit

littlexiang:
@msmeissn thx a lot. could you pls tell me the firmware version of your camera?
I would try to reset my repo master and make a clean PR, maybe squash all the commits into one

msmeissn:
it show 1.0.0 as firmware version.

littlexiang:
FYI my D1300 (v1.1.0) emits both events when single meta info change happens, so this event will be triggered twice

PR #313, opened 2018-10-06
Add VID:PID for Canon EOS 4000D using the same config as 1000D

knro:
So the topic is pretty much it.


Blueshawk:
Two of my favorite developers in one place. :dango: 
You guys rock. :)

PR #320, opened 2018-10-29
doc-tool update package name

royord:
package name updated from gtk-doc to gtk-doc-tools, this is the debian package name.


royord:
Thanks for the consideration.

ndim:
The package is also called `gtk-doc` on Fedora.

ndim:
Upstream also calls it `gtk-doc`.

PR #348, opened 2019-01-11
Fixed compiler error when using #warning on Visual Studio compiler

vividos:
When compiling libgphoto2 using Visual Studio, in one place the compiler chokes on the #warning statement in gphoto2-port-info-list.c. This PR fixes that by checking for _MSC_VER (which should be only set for Visual Studio and maybe MinGW) and uses #pragma warning instead.


vividos:
Thanks!

PR #349, opened 2019-01-12
Fixed including Win32 shlobj.h header when compiling using Visual Studio

vividos:
Here's another PR; gphoto2-port-portability.h undefines the "interface" definition, since "interface" is used throughout libgphoto2, but the Shlobj.h included from gphoto2-setting.c relies on this define. This all only occurs when compiling using Visual Studio and the Windows SDK 8.1 only. I guess MinGW doesn't have this define.


vividos:
Thanks again!

PR #362, opened 2019-01-29
ptp2: Correctly decode and set cRAW image format for Canon EOS

nowrep:
Since at least Canon EOS M50, image size and compression does not
fully describe used image format.
This patch encodes RAW flag (0x8) into image compression value
to correctly decode/encode image format.


msmeissn:
thanks!

PR #365, opened 2019-01-30
Add Fujifilm GFX 50R

seth-cohen:
Add Fujifilm GFX 50R usb ids to models and property summary txt
file


msmeissn:
could you move it 1 entry up, so the product ids are ascending? (If not I can do it)

you could also hide/blank the serial number from your summary output to avoid privacy issues

seth-cohen:
sorry yes... I will move it up.

PR #387, opened 2019-03-12
Added ID for Canon EOS 2000D

muesli:
I'm not entirely sure about requiring `PTPBUG_DELETE_SENDS_EVENT`, but I've added it since most other EOS models seem to use that flag. Is there some way I can test this?


msmeissn:
The flag might not be needed anymore if you use libusb1 , as we catch interrupts asynchronously.

it would be visible with libusb0 and deletion of several files , with e.g. gphoto2 -D -R 
and cause hangs during delete

PR #389, opened 2019-03-13
Add Canon EOS Rebel T6 VID:PID

knro:
As the title says.


msmeissn:
this looks kind of broken, lots of commits instead of just 1?

msmeissn:
can you also order the new entry so that the vendor id is linear increasing, so a bit lower ?

PR #393, opened 2019-03-16
Added PTP-IP support for Canon G7X "MK 1"

SparkyCola:
Successfully tested remote capture, remote file system, remote viewfinder.
However it looks like the change in  camlibs/ptp2/ptp.h is accidental and should be ignored.


msmeissn:
i merged some things, like the ID setting, the camera/ text.
i fixed some other pieces in place.

The setremotemode wraps I did not fix differently yet, neither the button==0 code

PR #415, opened 2019-05-13
Reduce Sony PTP capture time in manual focus and on RX0 in automatic focus

romainreignier:
Until now, in case of manual focus, the d213 property will not change so we were
waiting for the 1 sec timeout. So I propose to check the current focus mode before
to skip the useless wait.

On the Sony DSC-RX0, when the focus is ready, the d213 property becomes
3 and not 2 so the timeout where always reached as you can see in the attached logs file.
So I check that the value of `0x213` is either 2 or 3.

[trigger_capture_logs.txt](https://github.com/gphoto/libgphoto2/files/3173420/trigger_capture_logs.txt)


romainreignier:
I try to reduce the capture time because I was to take pictures as fast as possible.
I have tried to use the burst mode but the retrieving of the files is too slow. Do you have any tips to achieve 3-4 fps with libgphoto?
Thanks

Issue #426, opened 2019-07-03
Electronic shutter with Nikon Z6 throws -1 on gp_camera_capture


avisecag:
**Describe the bug**

I am trying to capture an image with the camera's silent mode (electronic shutter). Related camera widget is _0xD1AD_. I am not able to get the filepath on the camera to grab the image and gp_camera_capture throws error _-1_. This only happens with the silent mode turned on. In the commandline tool I get an unknown error but gphoto2 is still able to download the image from the camera.

**Name the camera**
Nikon Z6

**libgphoto2 and gphoto2 version**

gphoto2         2.5.23
libgphoto2      2.5.23
libgphoto2_port 0.12.0 

**To Reproduce**
[electronic_shutter.log](https://github.com/gphoto/libgphoto2/files/3353528/electronic_shutter.log) error -> line 900
[mechanical_shutter.log](https://github.com/gphoto/libgphoto2/files/3353529/mechanical_shutter.log)

avisecag:
I was able to resolve the bug:
in file camlibs/ptp2/library.c on line 3520
``C_PTP_REP (nikon_wait_busy (params, 100, 20000))``

must not be executed for Nikon Z 6 and probably Nikon Z 7 (can't test it at the moment)

``	if (params->deviceinfo.Model && (strcmp(params->deviceinfo.Model,"Z 6") != 0)) {
		C_PTP_REP (nikon_wait_busy (params, 100, 20000));
	}``

msmeissn:
What happens with this Busy wait code? is it reporting an error and which one?

msmeissn:
i see the logfs, sorry ... will check them

msmeissn:
0xa201 is the error returned on busy wait

PR #427, opened 2019-07-03
Added exclusion for Nikon Z6 when capturing in silent mode

avisecag:
Pull request to solve reported bug #426 with nikon Z6 when capturing an image in silent mode


msmeissn:
i fixed it a bit differently, i ignore the return code in the busy wait

Issue #444, opened 2019-09-14
/usr/include/gphoto2/gphoto2 infinite loop


hwpfeil:
**Describe the bug**

libgphoto2 installs /usr/include/gphoto2/gphoto2 which is a link pointing to './'

Infinite recursion. gphoto2-->gphoto2-->gphoto2....
Is there a compelling reason for this self-recursion?

**Name the camera**
Camera name as shown by gphoto2 --auto-detect or USB IDs
n/a
**libgphoto2 and gphoto2 version**
output of: gphoto2 --version

2.5.23

**To Reproduce**

ls -l /usr/include/gphoto2/gphoto2
ls -l /usr/include/gphoto2/gphoto2/gphoto2
ls -l /usr/include/gphoto2/gphoto2/gphoto2/gphoto2
...

When reproducing with the commandline tool, attach debugout using  --debug --debug-logfile=debug.log   if considered useful.

ndim:
This was introduced for source compatibility between the two forms

    #include <gphoto2.h>
    #include <gphoto2/gphoto2.h>

Of the top of my head, I cannot remember the exact scenario where this was introduced, I am looking at "git blame" and `git show fd692a071e` right now to help me remember.

What exactly is the problem with that symlink? Which program is exhibiting that "infinite recursion" behaviour?

ndim:
In September 2006, I made

     #include <gphoto2/gphoto2-foo.h>

the consistent way to include `libgphoto2_port` and `libgphoto2` headers (cf. https://github.com/gphoto/libgphoto2/commit/b8687e532ff267edd641064884e1cd506621fd1b). For compatibility with third party software using `libgphoto2` and still using `#include <gphoto2-foo.h>`, I added that symlink.

I just found that the Fedora package has been removing the symlink for their `libgphoto2` package since 2009 ([RHBZ #460807](https://bugzilla.redhat.com/show_bug.cgi?id=460807)) so there is probably no software around any more now (ten years later, in September 2019) which requires this.

So `libgphoto2` (and `libgphoto2_port`) could maybe drop the symlink at this time.

I still think if some program recursively traversing the filesystem tree (such as e.g. `grep -r` or `find`) can be confused by symlinks pointing somewhere, it is those programs which should be fixed. This is a different problem, though.

hwpfeil:
[...]

PR #452, opened 2019-10-17
Remove gphoto2 symlink loop in /usr/include/gphoto2

ndim:
Stop creating the gphoto2 symlink loop in /usr/include/gphoto2/.

The /usr/include/gphoto2/gphoto2 symlink pointing to `.`, i.e.
to /usr/include/gphoto2 was added in 2006 to solve an source
compatibility issue from back then. Fedora have been removing
that symlink from their packaged libgphoto2 packages since 2009,
so it appears those compatibility issues have been solved since
2009.

We should thus be able to remove the symlink now (in 2019).

Resolves: https://github.com/gphoto/libgphoto2/issues/444


msmeissn:
ok for me.

PR #498, opened 2020-05-05
Added support for ax203 based keychain

aaronw2:
This keychain had an unsupported flash as well as an unexpected
value in the expected string.  Once the driver accepts that unexpected
value it seems to work.

Signed-off-by: Aaron Williams <aaron.w2@gmail.com>

### DIFF
--- a/camlibs/ax203/ax203.c
+++ b/camlibs/ax203/ax203.c
@@ -13,7 +13,7 @@
  * GNU Lesser General Public License for more details.
  *
  * You should have received a copy of the GNU Lesser General Public License
- * along with this program; if not, write to the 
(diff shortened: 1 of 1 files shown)


jwrdegoede:
Thank you for your pull-req for this. There are really 3 sets of changes in here:

1. It seems that your editor has removed trailing whitespace in various places, that is good, but that really should be a in a cleanup commit by itself
2. The addition of the new eeprom model
3. Accepting buf[7] containing 0x52 in the parameter block validation code

Ideally this would be split out in 3 separate commits. Also please add a comment to the 0x52 check that some newer AX206-3.5.x models have 0x52 in byte 7.

PR #507, opened 2020-05-21
Remove duplicate Sony A7s entry with reduced functionality

trylle:
Was surprised to find that I had lost capture preview functionality on my A7s between versions 2.5.23 and 2.5.24. Tracked the regression down to f61a9d965. This PR removes the duplicate and reactivates the original entry at line 1057.


msmeissn:
Thanks! I moved the correct entry a bit higher to keep in sort order in a follup commit

PR #527, opened 2020-06-22
Speed up Sony shutterspeed changes

uchrisu:
This change implements larger steps when changing the shutterspeed. Before the values were changed by the smallest steps possible. Now the the "jump" is calculated according to the distance between current and new value. 
Not all Sony cameras reach the goal. Therefore, we will still need the loop.
Example: Sony a6400 from 30s to 1/4000s took over 40s before and takes now 2.5 seconds.
Even better: For example, Sony a9 can do all changes with only the initial step, so, reaching the goal instantaneously.

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -4285,7 +4285,7 @@ _get_Sony_ShutterSpeed(CONFIG_GET_ARGS) {
 
 static int
 _put_Sony_ShutterSpeed(CONFIG_PUT_ARGS) {
-	int			x,y,a,b,direction,num_items,position_current,position_new;
+	int			x,y,a,b,direction,position_current,position_new;
 	const char		*val;
 	float 			old,new,current;
 	PTPPropertyValue	value;
@@ -4293,8 +4293,6 @@ _put_Sony_ShutterSpeed(CONFIG_PUT_ARGS) {
 	PTPParams		*params = &(camera->pl->params);
 	GPContext 		*context = ((PTPData *) params->data)->context;
 	time_t			start,end;
-	
-	num_items = sizeof(sony_shuttertable)/sizeof(sony_shuttertable[0]);
 
 	CR (gp_widget_get_value (widget, &val));
 
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -4285,7 +4285,7 @@ _get_Sony_ShutterSpeed(CONFIG_GET_ARGS) {
 
 static int
 _put_Sony_ShutterSpeed(CONFIG_PUT_ARGS) {
-	int			x,y,a,b,direction;
+	int			x,y,a,b,direction,num_items,position_current,position_new;
 	const char		*val;
 	float 			old,new,current;
 	PTPPropertyValue	value;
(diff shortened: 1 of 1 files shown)


chubun:
Can this be done on other configurations such as f-number?  By the way, using the SDK from Sony (only Windows version at this time), these values seem to be set in one step on an alpha 7R4.

    On Monday, June 22, 2020, 3:37:17 PM CDT, uchrisu <notifications@github.com> wrote:  
 
 
This change implements larger steps when changing the shutterspeed. Before the values were changed by the smallest steps possible. Now the the "jump" is calculated according to the distance between current and new value.
Not all Sony cameras reach the goal. Therefore, we will still need the loop.
Example: Sony a6400 from 30s to 1/4000s took over 40s before and takes now 2.5 seconds.
Even better: For example, Sony a9 can do all changes with only the initial step, so, reaching the goal instantaneously.

You can view, comment on, or merge this pull request online at:

  https://github.com/gphoto/libgphoto2/pull/527

Commit Summary
   
   - Faster change of shutterspeed
   - num_items was not used

File Changes
   
   -  M camlibs/ptp2/config.c (48) 

Patch Links:
   
   - https://github.com/gphoto/libgphoto2/pull/527.patch
   - https://github.com/gphoto/libgphoto2/pull/527.diff

uchrisu:
It should be possible for f-number, iso, exposure-compensation. However, the implementation would need a little more effort in gphoto2 since they all use the macro PUT_SONY_VALUE and the implementations now have to be more individual.

I will create a pull request with fast changes of these properties, just need some more time for implementing it ;-)

I can only test on Sony A7III, A9 and A6400. 
The new generation (fourth) might be even more easier. However, also A9 shows different behavior to the other two since it is the only camera which reaches the goals instantaneous. For A7III and A6400 the step widths are only upper limits.

msmeissn:
Thanks! 
I have to try on my older SLT A58 too here.

For other values, it just the amount of table entries to jump over I would say, this seems doable in the generic setter too

msmeissn:
Perhaps something like:

if (posnew > posorig)
propval.u8 = posnew-posorig;
else
propval.u8 = 0x100 - (posorig-posnew);

uchrisu:
Please let me hear if it also helps improving the speed of the SLT A58 settings. 

I will do a fix for the other values this week. Probably I will extend the macro with two new arguments such as "list_type" and "list", since we need access to the list within the macro.

msmeissn:
The list is actualy the enum values.I will add a patch you can try.

msmeissn:
can you try current git with above patch?

uchrisu:
Oh I did not had time for a deeper look before, but you are right, it seems that everything what we need is already there. However, in the result your patch seems to work only in few cases on the real device (A9, A6400). Probably the problem is not the patch but something elsewhere. I will try to figure it out and get it to work as intended. May need some days.

Thanks!

msmeissn:
perhaps iot only works for some of the properties?

uchrisu:
I tried iso at first. The A6400 does make only single steps and the a9 is able to make the full jump. However, the jump is only done the first time I attached the camera. Afterwards dpd.FORM.Enum.NumberOfValues is zero and therefore the fallback mode with single steps is used. At the moment I don't know why and I try to figure it out, but the code is quite hard to follow because of all those pre-processor macros.

msmeissn:
Btw, I now have code that sets shutterspeed, iso and exposurebiascompensation directly on the "newer" sony a cameras like A7r4.

(These properties can now be set directly.)

PR #533, opened 2020-06-27
Add Docupen camlib

ondrej-zary:
Add camlib for Docupen RC800 wand scanner. Based on Docupen tools by Florian Heinz (https://sourceforge.net/projects/docupen/) and much improved (image correction according to calibration data from scanner, greyscale support).


msmeissn:
docupen/README.docupen was not there, I commented out thre Makefile-files entry

PR #545, opened 2020-07-27
Remove trailing whitespace from all *.h and *.c files.

sh0:
Is there any chance we could get rid of trailing whitespace in source files? I know it is a fairly cosmetic change and the diff is huge, but I think it would help. Some subjective reasons for this change could be found [here](https://softwareengineering.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal) and [here](https://stackoverflow.com/questions/300489/why-is-it-bad-to-commit-lines-with-trailing-whitespace-into-source-control).


msmeissn:
makes futuree git blame kind of hard though. Need to think about it.

sh0:
Yeah, that is true.

Most of the activity in libgphoto2 seems to happen around `ptp2` and `libusb` and few other places. One option is to avoid changing those parts where there has been considerable activity in last year or two. Most of this commit is actually camlibs that haven't changed much in the last 10 years.

In the end it is not a big deal if this commit doesn't get merged. If you'd like to see some changes then just let me know.

PR #548, opened 2020-08-03
Fix various issues to reduce compiler warnings

sh0:
This pull request consists of 44 commits to different camlibs, iolibs, core code and examples to reduce number of warnings given by compilers.

The motivation is to be able to compile the whole codebase with `-Wall -Wextra` without being overwhelmed by compiler output. While most of the edits submitted here had no real impact, there were number of bugs already found. Each commit has a summary of changes.

Code was tested on GCC 8.4.0, 9.3.0 and 10.1.0 with `-Wall -Wextra -Wno-unused-parameter -Wno-format-overflow -Wno-format-truncation -Wno-stringop-truncation` and on Clang 10.0.0 with `-Wall -Wextra -Wno-unused-parameter`. Currently only `x86_64` architecture was used.

I don't know how much of these changes are acceptable, but let me know if something should be changed or submitted as separate PR.

### DIFF
--- a/camlibs/agfa-cl20/agfa_cl20.c
+++ b/camlibs/agfa-cl20/agfa_cl20.c
@@ -153,7 +153,7 @@ get_file_func (CameraFilesystem *fs, const char *folder, const char *filename,
 {
 	Camera *camera = privdata;
 	int n = -1;
-	int size = -1;
+	unsigned int size = -1;
 	unsigned char hb, lb;
 	unsigned long j;
 	unsigned int app1len = -1;
--- a/camlibs/ax203/ax203.c
+++ b/camlibs/ax203/ax203.c
@@ -45,69 +45,69 @@ static const struct eeprom_info {
 	int has_4k_sectors;
 	int pp_64k;
 } ax203_eeprom_info[] = {
-	{ "AMIC A25L040", 0x37133037,  524288, 1 },
-	{ "AMIC A25L080", 0x37143037, 1048576, 1 },
-	{ "AMIC A25L40P", 0x1320377f,  524288, 0 },
-	{ "AMIC A25L80P", 0x1420377f, 1048576, 0 },
-	{ "AMIC A25L16P", 0x1520377f, 2097152, 0 },
+	{ "AMIC A25L040", 0x37133037,  524288, 1, 0 },
+	{ "AMIC A25L080", 0x37143037, 1048576, 1, 0 },
+	{ "AMIC A25L40P", 0x1320377f,  524288, 0, 0 },
+	{ "AMIC A25L80P", 0x1420377f, 1048576, 0, 0 },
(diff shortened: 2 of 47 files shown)


msmeissn:
These kind of big patchsets make it hard to review.

I am  cherrypicking "good" commits piece by piece.

In general you have an antipattern:

-  "if (x < y)" to "if (x < (int)y)"  almost always risky, as it can introduce more problems (you now have to handle negative integers).

-  If you need to use casts, usually its risky.

Both would need to be very carefully reviewed or even tested.

msmeissn:
I did a first pass and cherry-picked the one that seemed good. (your branch should still be rebasable.)

msmeissn:
If you do multiple different fixes per camlib, it would be good to have them in separate commits.

the ptp2 change is too big at once, I carefully need to go over it.
the pentax change ... everything but library.c is copied 1:1 from pktriggercord. We need to see what we can merge back to them.

sh0:
Thanks for taking a look at this. I didn't want to submit ton of PRs without some feedback first so I had everything bundled together. I'll rebase and try to reorganize things per your comments.

About that `if (x < (int)y)` pattern. I don't like casting either. In many places I could change the type of some variables safely which would solve the issue. However sometimes it would have required a lot of changes or eventually some cast anyways so I went with a cast in comparison. I'll see if I can do more in that regard. I don't think that current solution is particularly good either - any small negative signed integers in those signed-unsigned compares end up huge unsigned integers.

msmeissn:
yeah, i saw I could cherry pick nicely.

All of those "cleanups" are a bit risky if we are not testing it.
I have some of the cameras also for older drivers, but mostly not. :/

Especially for ptp2 and some I have cameras for i probably can try the warnings myself and figure them out step by step at least.

sh0:
I have now edited most of the commits and rebased them on master. The code behavior was kept exactly the same whenever possible. E.g signed-unsigned compares would be kept unsigned per C standard if the arguments were with same size. I decreased the number of casts as much as possible. In some cases there are, however, casts to unsigned types. These casts would typically happen anyways implicitly. Finally I reviewed everything once more and wrote longer explanations for the changes in commit messages.

Missing changes:
* Pentax changes have been sent to pktriggercord project as PRs. Looks like it is going to take a while. I will post a separate PR when they are done.
* PTP2 changes will be posted as separate PRs.

I can submit separate PRs if requested, but currently kept everything here in one package for cherry-picking.

msmeissn:
i merged it manually. I also added the gcc options as default.

my tree is now largely free of these warnings (2 or 3 trickier ones still there)

sh0:
Thanks a lot for dealing with all the things that I sent in.

Bunch of things have been upstreamed to pktriggercord now, but one last exposure mode bugfix is remaining.

Issue #409, opened 2019-04-23
Lumix HTTP/Wifi protocol


totoantibes:
Fill in fields below. If you don't know how, leave it free.

**Name of the camera*
All Lumix G, GX, GH series. (I can test on G80 and GH4)
seems that only GH4, GX80 and GH5 have a usb thethered mode available. Many other Lumix do not offer PTP tethered camera capture/liveview etc. Only option is to use an HTTP protocol over Wifi. (or maybe BLE for newer cams) there are many commands available via this protocol such as shutter speed, iso, aperture, quality, focus, zoom. rec start/stop, capture and capture cancel. also the images are avail to download to in RAW, JPG and Thumbs. A VGA/QVGA live view data stream is also available. 

There are various running examples of programs that do take advantage of this protocol  and woudl be great to open Gphoto2/ligphoto2 to Lumix cameras. Unfortunately the HTTP/Wifi connectivity is a bit of an issue to plug into the camlib/port structure of libgphoto as i understand it from a very high level.

here is a list of github availalble programs that take advantage of this protocol
-https://github.com/palmdalian/python_lumix_control 
- https://github.com/peci1/lumix-link-desktop (full app in java)
-https://github.com/Ultrawipf/qtpanaremote (simple QT based app in C)
-https://github.com/totoantibes/LumixCameraAscomDriver (my own Ascom driver for astrophotography written in VB.NET) it does implement the picture download of the last capture over http/get with retry after parsing the XML of the available pics. 
  
Use the Marketing name.

**USB IDs**
e.g. by running:   lsusb 

**camera summary output**
run

  gphoto2 --summary > summary.txt

attach summary.txt here

**camera configuration output**

gphoto2 --list-all-config > list-all-config.txt

and attach list-all-config.txt here

**test capture**

test if capture perhaps already works:

  gphoto2 --capture-image-and-download
 
  gphoto2 --capture-preview

I am  a keen astrophotographer and would really like to use the libgphoto lib to interface with astrophoto session SW like ekos/kstars to control the Lumix CAMs. 

If you have no time/priority to dev this new family of cameras I woudl be keen to do it myself but some quidance/support would be needed prior to hacking something in this complex body of work

msmeissn:
This is something interesting to pursue :/ 
it needs a new camlib, perhaps a new port driver (or reuse ptpip somehow)

totoantibes:
well,

as a matter of  fact i started creating a new camlib (called lumix.c) and attached here. however 
[...]

PR #558, opened 2020-08-28
lumix: enable all video types, do not set quality

willat8:
Relates to #409.
I've been using these fixes for a few months and think it's a better default behaviour.

1. Don't change the user-set quality mode
2. Only call `switchToRecMode` when starting a livestream, not continuously throughout
3. Include all AVC_MP4 videos in file listing


msmeissn:
sounds good

PR #565, opened 2020-09-25
Added Nikon Z5

avisecag:
Added Nikon Z5 USB is and camera summary nikon-z5.txt


msmeissn:
I moved the ID to be in the correct sorted order after your commit.

PR #573, opened 2020-10-24
Not only Linux has statfs()

wferi:
This makes gphoto2 test031 pass on GNU/Hurd.

And maybe on kFreeBSD as well, one time. On the other hand, I've got no idea why the Linux constraint was put here originally, does this break the Windows builds for example? I can't readily test that.


msmeissn:
i took a parralell submission  in https://github.com/gphoto/libgphoto2/pull/572

PR #578, opened 2020-11-10
Add support for reading Canon EOS MovieMode switch and setting it on/off

### LINKED ISSUE #574: Switching between movie and photo modes (Canon)
Hi,
I've been using gphoto for several years, but have never had to chance to say thanks for this amazing project; Thanks!

One thing that I haven't been able to do, is programmatically switch between movie and photo modes on Canon cameras. Currently, you can start and stop movie recording using gphoto, but the camera has to be switched physically to movie mode.

Recently I noticed that this functionality is possible in Canon's EOS Utility, without having to use the physical switch. I checked Canon's EDSDK, and it does describe how to do this, using two commands; MovieSelectSwON and MovieSelectSwOFF.

I think this would be a really useful feature to have in gphoto. Would the values in the EDSDK be any help in implementing this, or would the commands need to be reverse engineered?

Henry

trilader:
This adds support for reading the state of the hardware "Movie Mode" switch setting on Canon EOS cameras as well as switching between still image and movie capture mode in software, without needing to move said switch.

Originally I wanted to implement reading the switch state and setting the mode as a single configuration item but couldn't figure out how to read the device property the regular way and not issue SetDevicePropertyEx commands to set it but only use a custom function.

Also I'm not sure if `_get_Canon_EOS_MovieModeSw` is a good implementation of a "write only"/"action trigger" configuration item (though other implementations nearby do exactly that).

I think this fixes #574

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -7607,6 +7607,30 @@ _put_Generic_OPCode(CONFIG_PUT_ARGS)
 }
 
 static int
+_get_Canon_EOS_MovieModeSw(CONFIG_GET_ARGS) {
+	int val;
+
+	gp_widget_new (GP_WIDGET_TOGGLE, _(menu->label), widget);
+	gp_widget_set_name (*widget,menu->name);
+	val = 2; /* always changed */
+	gp_widget_set_value  (*widget, &val);
+	return (GP_OK);
+}
+
+static int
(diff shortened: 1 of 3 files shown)


msmeissn:
looks good

if you cant read out the moviemode, the trigger is ok. If it can be read out, we need to do another methiod

PR #596, opened 2020-12-04
WIP: Make PTP/IP implementation work on Windows

trilader:
This PR makes the PTP/IP implementation (regular and Fuji) work on Windows.

Most of the work was wrapping the platform specific quirks/differences between WinSock2 and regular BSD sockets in some macros which abstract away the differences. This also sends the correct hostname to the camera instead of the fixed `gpwindows` string.

- The Fuji implementation is untested as I don't have a Fuji camera.
- The regular implementation was tested with a Canon EOS 77D.

**What I still need help with:**
My autoconf/automake skills are basically nonexistent. The ptp2 camlib need to link to `ws2_32.lib` when building on Windows. For development I worked around that by running `./configure LDFLAGS="-lws_32" ...` but that's not something that should be expected for people to do explicitly.

**Point for discussion:**
WinSock2 needs to be initialized before use by calling `WSAStartup`. I'm not sure who (application using libgphoto vs. libgphoto2 itself) should do that. I think having the application do it makes more sense as the application might also needs to do other network stuff for which that could be needed anyway.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -63,6 +63,10 @@
 #include "ptp-pack.c"
 #include "olympus-wrap.h"
 
+#ifdef HAVE_LIBWS232
+#include <winsock2.h>
+#endif
+
 #define USB_START_TIMEOUT 8000
 #define USB_CANON_START_TIMEOUT 1500	/* 1.5 seconds (0.5 was too low) */
 #define USB_NORMAL_TIMEOUT 20000
@@ -3081,6 +3085,11 @@ camera_exit (Camera *camera, GPContext *context)
 		gp_port_usb_clear_halt
 				(camera->port, GP_PORT_USB_ENDPOINT_INT);
 	}
+#if defined(HAVE_LIBWS232) && defined(WIN32)
+	else if ((camera->port!=NULL) && camera->port->type != GP_PORT_PTPIP) {
+		WSACleanup();
+	}
+#endif
 	return (GP_OK);
 }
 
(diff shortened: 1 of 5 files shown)


msmeissn:
i think your other change I just merged caused a conflict, you will need to rebase.

In general this approach looks good from my side. 
I have no idea where to put WSAStartup easily ... :/

trilader:
I originally developed this on top of the other change but submitted it separately so the PRs could be applied in any order. I've now rebased it to also include the other commit.

Do you have any ideas on how to do the autoconf thing?

trilader:
Regarding `WSAStartup`: According to [MS Docs](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsastartup) it is okay to call it multiple times as long as you call `WSACleanup` the same number of times (they've got some internal ref-counting going on).

This would make it viable to call e.g. in `gp_context_new()` and run the clean up when that context is destroyed.
That way we can make sure it's initialized and the embedding application doesn't **have** to take care of it.

msmeissn:
I added a configure check for ws2_32 now, can you try if that works?

trilader:
It now builds without `LDFLAGS="-lws_32"` after fixing what looks like a copy/paste mistake: The new configure check looks at `x$with_jpeg` instead of `x$with_ws232`. I noticed because the build I'm usually using doesn't use libjpeg.

I've also added a call to `WSAStartup` in camera_init and `WSACleanup` in camera_exit. This should be save as the docs say it can be called multiple times as long as the number of Startup/Cleanup calls are the same.

trilader:
Also sorry for the force pushes. I commited from a different computer and had my git author not setup correctly.

msmeissn:
No worries. I adjusted configure.ac to replace the jpeg -> ws232 typo.

I think you will need to include winsock.h also to library,c to avoid implicit function definition warnings?

trilader:
Good point. I've added that.

camlibs/ptp2/library.c
```diff
@@ -3081,6 +3085,11 @@ camera_exit (Camera *camera, GPContext *context)
 		gp_port_usb_clear_halt
 				(camera->port, GP_PORT_USB_ENDPOINT_INT);
 	}
+#if defined(HAVE_LIBWS232) && defined(WIN23)
```
msmeissn, line 3088: WIN23 or WIN32? (typo?)

trilader: That was indeed a typo - it is now fixed.

msmeissn:
I merged it as is... we can do improvements on top :)

PR #605, opened 2020-12-31
Recognize Nikon Z 7II and fix VendorExtensionID over PTP/IP

fahrion2600:
This PR adds the vendor and product code for the Nikon Z 7II. It also fixes the VendorExtensionID for the Z 7II (and probably other Nikon cameras) when connecting to the camera over PTP/IP, such as via the built-in Wi-Fi.

**Tested:**
* auto-detect shows "Nikon Z7_2"
* summary command displays the correct Vendor Extension ID "0xa" for Nikon and shows additional configs
* Capture and image download both work over USB and Wi-Fi


msmeissn:
ok

PR #625, opened 2021-02-21
Remove stdint.h shim

Zeranoe:
stdint.h has wide adoption at this point, so it should be safe to use.


ndim:
No argument about normal post 2010 Linux systems from me.

Some people are building for weird systems, though.

msmeissn:
i wonder since how long this stdint.h is present... it seems standardized in 1999? was it already shipped then?

msmeissn:
/usr/include/stdint.h /* Copyright (C) 1997,1998,1999,2000,2001,2006 Free Software Foundation, Inc.
at least in glibc

msmeissn:
also see stdint.h in UnitedLinux 1.0 from 2002.

msmeissn:
merged. we will see if someone reports issues.

ndim:
stdint.h came with C99.

PR #626, opened 2021-02-23
Fix byte swap calls

Zeranoe:
gphoto2-endian.h provides shims for the standard byte swap macros, so
using the standard name is preferred.


msmeissn:
thanks!

PR #627, opened 2021-02-26
Add model detection for Fujifilm X100V

gergelyerdelyi:
Starting with firmware version 2.0 the camera supports movie
capture in the X WEBCAM USB connection mode.


msmeissn:
thx!

PR #637, opened 2021-03-16
PTP/IP: Better error reporting and timeout handling

trilader:
The current PTP/IP implementation doesn't have very granular error reporting and returns PTP_RC_GeneralError where a more specific error cause is available. Another issue in situations with flakey and/or suboptimal WiFi coverage is that libgphoto can get stuck on blocking read(), write() or connect() calls when trying to talk to the camera for an undefined amount of time which causes the application using libgphoto to hang (or at least its gphoto thread).

This PR addresses both issues be introducing non blocking socket based communication with timeout as well as adding/adjusting the error codes returned.

This code was tested on Linux, Windows and macOS (mainly with Canon cameras).


msmeissn:
thanks!

PR #639, opened 2021-03-19
Arkell fix r5 aperture

msmeissn:
merge over special EOS R5 aperture


msmeissn:
thx, i merged into libgphoto2 master

PR #643, opened 2021-03-31
Remove obsolete gp2ddb (gphoto2 device database) idea

ndim:
The gphoto2 device database (gp2ddb) was originally intended to
replace the device lists in the camlib code by a device list
maintained in a text file outside of the camlib C code.

This device list file could then be changed without
recompiling camlibs, at the cost of adding a simple gp2ddb
to internal binary representation compiler to libgphoto2.

This would have allowed a user to very easily add a new camera
model to the list of camera models supported by a specific
camlib, as long as that new camera model uses the same
protocol as already supported camera models do.

As the gp2ddb idea has never been developed any further in the
15 yearns since my initial commit from 2006, this idea appears
to be obsolete and can thus be removed from the codebase.


ndim:
I created this pull request instead of just pushing this to allow for a discussion, if there happens to be any interest after all.

msmeissn:
With the ptp2 driver supporting new cameras of the same class okayish currently, and/or new models needing code changes potentially, i think we do not need this.

good to remove for me.

PR #649, opened 2021-04-19
camlibs/ptp2: Unpack Canon EOS changes for INT32 properties

trilader:
White implementing a control for the white balance shift I noticed the WhiteBalanceAdjust{A,B} properties where always reported as 0.
After investigating a bit I noticed that changes for `PTP_DTC_INT32` properties weren't being unpacked which resulted in missed updates when the value changes on the camera and libgphoto getting out of sync.


msmeissn:
thx!

Issue #653, opened 2021-04-21
Fuji X-T4: Local camera control / gp_camera_trigger_capture


trilader:
**Describe the bug**

I'm not sure if this is actually a bug report or either a feature request, development advice for a patch for me to submit or me just using libgphoto wrongly. By default libgphoto switches the camera to USB control which disables all local controls on the camera.

**What I want to do**
I want to have a user of a software be able to locally control the camera (press the trigger, use the remote release via the 2,5mm TRS jack, change settings on the camera). As well as programatically trigger the camera from the software using `gp_camera_trigger_capture`.

**What I've tried**
I found out that `camera_prepare_capture` DPC 0xd207 gets set to 2 which is USB mode. By default it's set to 1 which I assume is camera mode. It also gets set back to that in `camera_unprepare_capture`. When I remove the setting of it to 2 in `camera_prepare_capture` the local camera controls work as expected and when the camera gets triggered either by the remote shutter release input or the shutter button the camera takes a picture, libgphoto detects it and the software downloads it to the computer.

What doesn't work without USB control mode is `gp_camera_trigger_capture`. In it, after fixing the type of the wait for propval 0xd212 loop from UINT64 to UINT16, the `ptp_initiatecapture ` after triggering the focus action fails with `'PTP Device Busy' (0x2019)`.

I've tried to work around that by setting 0xd207 to 2 before focusing the camera, which works and allows the rest of `gp_camera_trigger_capture` to work normally, and then resetting it at the end so the user can control the camera again, which failed with `'PTP Device Busy' (0x2019)`.

My current idea is that the issue with changing the mode back is because the camera just took an image and isn't ready to switch back before it was transferred/removed as the mode can be changed back in unprepare_capture after the image was removed from the camera.

It that is the case I think something like the uilock config option for Canon camera might be a good idea.

**Name the camera**
`Fuji Fujifilm X-T4             usb:002,029`
`Bus 002 Device 029: ID 04cb:02e6 Fuji Photo Film Co., Ltd USB PTP Camera`

**libgphoto2 and gphoto2 version**
I'm using libgphoto2 on the master branch based of ff6013b.

**To Reproduce**
I'm using a simple test program which opens the camera, starts polling events for 5 seconds and after 1 second of that calls `gp_camera_trigger_capture` and after the loop exits the camera. I can provide if it you think it's helpful.

trilader:
I've investigated some more and made some progress.

[...]

PR #654, opened 2021-04-22
Misc Fuji fixes and improvements

trilader:
While investigating #653 I fixed some bugs and added a new config option for Fuji cameras.

In this case this is:
- A fix for a wrong property data type in `camera_trigger_capture`
- A new config option to change camera priority mode (USB remote or local camera control)
- Replaced usage of magic numbers with the corresponding DPC #define
- A fix for active live-view (via repeatedly calling `camera_capture_preview`) conflicting with tethered shooting by downloading new images when detected by `camera_wait_for_event`

This is only tested with a Fuji X-T4 as that is what I have access to but I don't think anything should break with other Fuji cameras.


msmeissn:
looks good, thanks!

PR #661, opened 2021-05-15
sierra: Fix Epson PhotoPC 500

ondrej-zary:
Got an old Epson PtohoPC 500 camera only to find that it does not work with gphoto2.
Seems it has problems with speed over 38400bps, but it works fine in Windows at 115200bps.
Add the same workaround that Windows software uses - write data one byte at a time if the first command times out.
Also found that the camera responds with unknown 0x18 response to the "folder test" command.

### DIFF
--- a/camlibs/sierra/library.c
+++ b/camlibs/sierra/library.c
@@ -464,8 +464,9 @@ sierra_write_packet (Camera *camera, char *packet, GPContext *context)
 		if (camera->pl->flags & SIERRA_NO_BLOCK_WRITE) {
 			for (x = 0; x < length; x++)
 				CHECK (gp_port_write(camera->port, &packet[x], 1));
-		} else
+		} else {
 			CHECK (gp_port_write (camera->port, packet, length));
+		}
 	}
 
 	return GP_OK;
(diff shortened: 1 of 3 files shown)


msmeissn:
cancel patch is ok, other patch i need to look a biut into it

camlibs/sierra/library.c
```diff
@@ -460,7 +461,11 @@ sierra_write_packet (Camera *camera, char *packet, GPContext *context)
 			(camera->pl->flags & SIERRA_WRAP_USB_MASK),
 			packet, length));
 	} else {
-		CHECK (gp_port_write (camera->port, packet, length));
+		if (camera->pl->flags & SIERRA_NO_BLOCK_WRITE) {
+			for (x = 0; x < length; x++)
+				CHECK (gp_port_write(camera->port, &packet[x], 1));
+		} else
+			CHECK (gp_port_write (camera->port, packet, length));
```
hfiguiere, line 468: ultimately it's Marcus call but I find the 'optional' curly braces confusing. At least the `else` block should be in curly braces too.

msmeissn:
thanks, i merged it. will need to see if i still have a serial sierra camera here for own testing ;)

PR #672, opened 2021-06-29
Fixed an error check on _set_config

ben5516:
When updating multiple settings at the same time, this continue was allowing an early failure to be ignored (if the next setting returned GP_OK).

Not sure if this is the best route to fix.


msmeissn:
the check is for cases where we have multiple entries with the same "name" ... so changing it is probably not right, needto think about it

PR #675, opened 2021-07-09
Revamp Sony ISO MFNR logic

RReverser:
This PR contains various fixes for Multi-Frame Noise Reduction ISO mode in Sony cameras.

 - Add support for distinguishing Standard vs High modes for Multi Frame Noise Reduction.
 
   Previously those would be rendered as separate options with the same name, making it impossible to choose the correct one or switch from one to another.
 
   Now the high mode is indicated by the `+` suffix.
 - Fix higher ISO values for MFNR too.
 
   Previously a mask 0xffff was used, which was incorrect and didn't cover higher ISO values. As a result, they were rendered incorrectly.
   
   Instead, base ISO should be read with 0x00ff_ffff mask, and the highest byte then indicates the MFNR mode (0 for disabled, 1 for Standard, 2 for High). This also simplifies the logic for Auto ISO, since it's correctly covered by the updated mask.

I also went ahead and wrote a one-off script to update .txt files for Sony cameras to show the fixed values.


msmeissn:
looks goodl, thanks :)

PR #676, opened 2021-07-09
Fix infinite loop when setting Sony values

RReverser:
An example where I ran into this: Sony a6600 reports values [25, 50, 64, 80, 100, 125, 160, 200, 250, 320, 400, ...] as supported with Multi Frame Noise Reduction. However, among those values the camera only allows selecting between [100, 200, 400].

As a result, switching even between valid values - say, ISO 100 with MFNR and ISO 3200 without MFNR - was calculating incorrect offset, overshooting in one direction, calculating new incorrect offset, overshooting in another direction, and so on leading to an infinite loop.

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -1329,10 +1329,20 @@ fallback:										\
 			GP_LOG_D("posnew %d, posorig %d, value %d", posnew, posorig, value);	\
 			if (posnew == posorig)						\
 				break;							\
+			/*
+			Note: Sony seems to report values in an enum that are not
+			actually supported under certain configurations (e.g. ISO).
+			If we use `posnew - posorig` to ask camera to jump to the
+			required position, it will end up overshooting, and on the next
+			cycle will try to move by the same `posnew - posorig` in the
+			opposite direction, overshoot again, and so on infinitely.
+			Instead, we have to ask camera to +/- values one by one and read
+			back the value it _actually_ switched to each time.
+			*/ \
 			if (posnew > posorig)						\
-				propval.u8 = posnew-posorig;				\
+				propval.u8 = 0x01;				\
 			else								\
-				propval.u8 = 0x100-(posorig-posnew);			\
+				propval.u8 = 0xff;			\
 		} else {								\
 			if (value == origval)						\
 				break;							\


msmeissn:
this is actually code that should make switching values faster than usual.
i will adjust a bit afzter merging

RReverser:
> this is actually code that should make switching values faster than usual.
> i will adjust a bit afzter merging

Yeah I understand the purpose and it's sad to have to use +- one by one, but on practice the result seems to be the opposite.

I didn't observe any visible performance degradation for values within the valid ranges either.

msmeissn:
i added code now that tries the big jump first, then single steps afterwards.

i need to rent a sony again to debug this further i fear... not sure how to find it out correctly

RReverser:
Let me try this quickly...

RReverser:
Just tried. Your change does work, but rather inefficiently when user is making small adjustments over such ranges (which, I'm guessing is the most common usecase?).

Using the example from the description - if I want to go from 100 MSNR to 200 MSNR, then with linear +1 steps it would be just one step.

With the updated logic it will first calculate the offset as +3, ask the camera to do 3 steps, the camera will actually jump 100 -> 200 -> 400 -> 800, and now on the next iteration it will start the linear -1 steps, going 800 -> 400 -> 200.

As a result, we get 3 operations instead of 1.

I guess the "average" best option would be to implement a proper binary search, halving the offset on each step, but not sure if it's worth it...

RReverser:
Of course, an even better option would be to somehow get _actually supported_ list of values from Sony, but I'm not sure if it's possible.

msmeissn:
hmm, do you have any indication how many steps were necessary compared to how many gphoto calculated? just 1 ?

RReverser:
> hmm, do you have any indication how many steps were necessary compared to how many gphoto calculated? just 1 ?

For 100 -> 200 - just 1 step, yeah (whereas gphoto calculates as 3 because it thinks the camera will go over [100, 125, 160, 200]).

RReverser:
Although in other cases - for wider changes - your method does seem more efficient. Maybe binary search would indeed be the best option to cover all kinds of changes...

msmeissn:
its a bit weird that it reports those values in the enum, but does not step throuzgh them. is there perhaps another option that makes it more finegrained or so... :/

is the a6600 not supporting the direct setting of iso via property d226 ?

RReverser:
> is the a6600 not supporting the direct setting of iso via property d226 ?

Unfortunately, no.

> its a bit weird that it reports those values in the enum, but does not step throuzgh them.

Yeah initially when I ran into this, it turned out that none of the multi-frame noise reduction options are supported because I had a "picture style" set in settings. As a result, trying to set any of those values immediately ran into infinite loop.

My understanding is that they have some static list of values hardcoded in the software they report back, while the actual values depend on runtime config (and maybe even sensor itself?).

This is not limited to ISO either - for example, I see f-number configs going all the way down to `f/1`, even though the kit lens camera has on only goes as low as `f/3.5`.

RReverser:
Actually even for large steps the `firstloop` logic seems to result in a lot more operations due to overshooting that wraps around the enum.

Here's what going from 100 MFNR to 102400 MFNR looks like with this PR (just +1 steps):

```
value changed (0x10000c8 vs 0x1000064 vs target 0x1019000), next step....
value changed (0x1000190 vs 0x10000c8 vs target 0x1019000), next step....
value changed (0x1000320 vs 0x1000190 vs target 0x1019000), next step....
value changed (0x1000640 vs 0x1000320 vs target 0x1019000), next step....
value changed (0x1000c80 vs 0x1000640 vs target 0x1019000), next step....
value changed (0x1001900 vs 0x1000c80 vs target 0x1019000), next step....
value changed (0x1003200 vs 0x1001900 vs target 0x1019000), next step....
value changed (0x1006400 vs 0x1003200 vs target 0x1019000), next step....
value changed (0x100c800 vs 0x1006400 vs target 0x1019000), next step....
```

Here's what it looks like with the updated logic from https://github.com/gphoto/libgphoto2/commit/2ed0113e6fd6b76dd60ef019e9c299cf6150c196:

```
value changed (0xffffff vs 0x1000064 vs target 0x1019000), next step....
value changed (0x32 vs 0xffffff vs target 0x1019000), next step....
value changed (0x40 vs 0x32 vs target 0x1019000), next step....
value changed (0x50 vs 0x40 vs target 0x1019000), next step....
value changed (0x7d vs 0x50 vs target 0x1019000), next step....
value changed (0xc8 vs 0x7d vs target 0x1019000), next step....
value changed (0x140 vs 0xc8 vs target 0x1019000), next step....
value changed (0x1f4 vs 0x140 vs target 0x1019000), next step....
value changed (0x320 vs 0x1f4 vs target 0x1019000), next step....
value changed (0x4e2 vs 0x320 vs target 0x1019000), next step....
value changed (0x7d0 vs 0x4e2 vs target 0x1019000), next step....
value changed (0x9c4 vs 0x7d0 vs target 0x1019000), next step....
value changed (0xc80 vs 0x9c4 vs target 0x1019000), next step....
value changed (0xfa0 vs 0xc80 vs target 0x1019000), next step....
value changed (0x1388 vs 0xfa0 vs target 0x1019000), next step....
value changed (0x1900 vs 0x1388 vs target 0x1019000), next step....
value changed (0x1f40 vs 0x1900 vs target 0x1019000), next step....
value changed (0x2710 vs 0x1f40 vs target 0x1019000), next step....
value changed (0x3200 vs 0x2710 vs target 0x1019000), next step....
value changed (0x3e80 vs 0x3200 vs target 0x1019000), next step....
value changed (0x4e20 vs 0x3e80 vs target 0x1019000), next step....
value changed (0x6400 vs 0x4e20 vs target 0x1019000), next step....
value changed (0x7d00 vs 0x6400 vs target 0x1019000), next step....
value changed (0x9c40 vs 0x7d00 vs target 0x1019000), next step....
value changed (0xc800 vs 0x9c40 vs target 0x1019000), next step....
value changed (0xfa00 vs 0xc800 vs target 0x1019000), next step....
value changed (0x13880 vs 0xfa00 vs target 0x1019000), next step....
value changed (0x19000 vs 0x13880 vs target 0x1019000), next step....
value changed (0x1ffffff vs 0x19000 vs target 0x1019000), next step....
value changed (0x2000190 vs 0x1ffffff vs target 0x1019000), next step....
value changed (0x20000c8 vs 0x2000190 vs target 0x1019000), next step....
value changed (0x2000064 vs 0x20000c8 vs target 0x1019000), next step....
value changed (0x2ffffff vs 0x2000064 vs target 0x1019000), next step....
```

Result: 9 steps vs 33 steps.

RReverser:
Tried my hand at binary search, but only reintroduced the infinite loop instead. I guess for now we can keep it as it is.

Or maybe special-case the ISO config to always use +1/-1 steps, while the rest can attempt the proper offset first?

As far as I can tell, this is the only property that suffers from so many wrong values in the list.

msmeissn:
hmmm, yeah, hardcoded list is imaginable. can the camera set values between 100 and 200 itself? or do the intermediates exist on camera?

RReverser:
> 
> 
> hmmm, yeah, hardcoded list is imaginable. can the camera set values between 100 and 200 itself? or do the intermediates exist on camera?

Some of these intermediates exist and can be selected in non-MFNR mode, but not when multi-frame noise reduction is used. I imagine it also might differ from camera to camera.

PR #678, opened 2021-07-12
Fix overshoot logic for Sony enums

RReverser:
Overshoot logic was always comparing values to determine if the desired target has been overshot. However, in case of enums this didn't work correctly, since enum values are not guaranteed to be in the sorted order.

For example, "Auto ISO Multi Frame Noise Reduction" has a higher numerical value than, say, "ISO 400 Multi Frame Noise Reduction" but comes earlier in the enum list, so trying to go from "ISO 102400" to "ISO 400 Mutli Frame Noise Reduction" would stop at "Auto ISO Multi Frame Noise Reduction" thinking that the value has been overshot and can't go any further.

This PR fixes the logic to use enum index for overshot comparisons instead when `useenumorder` is enabled.

~~Additionally, the overshot check is skipped on the first loop where we make a large leap, since overshooting is pretty much inevitable in that case, and we should check for it only on small +-1 adjustments.~~

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -1309,11 +1309,13 @@ _put_sony_value_##bits (PTPParams*params, uint16_t prop, inttype value,int useen
 		return GP_OK;								\
 	}										\
 fallback:										\
+	useenumorder = useenumorder && (dpd.FormFlag & PTP_DPFF_Enumeration) && dpd.FORM.Enum.NumberOfValues; \
 	do {										\
+		int posorig = -1, posnew = -1, posnow = -1; \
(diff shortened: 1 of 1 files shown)


RReverser:
Wait just noticed that I'm checking `firstloop` after it's been reset to 0. I'll fix this...

RReverser:
Ah right it worked because actually `firstloop` check doesn't matter at all because `propval.u8` will not match `0x01`/`0xff` if we made a large jump anyway (and if we did make a small jump, it's ok to check for overshoot already).

I've updated the check and comment correspondingly.

PR #679, opened 2021-07-13
Mark properties without setters as readonly

RReverser:
Without this, widgets for things like "Serial Number", "Manufacturer", "Battery level" etc. would be created as writable, even though they are not.

It it possible to special case them in GUI frontends when rendering (e.g. by making everything under "status" section readonly anyway), but it seems better to return correct readonly status for such properties from libgphoto2 itself.


msmeissn:
ok, thanks!

PR #681, opened 2021-07-14
Normalize Sony shutter speeds

RReverser:
Apparently Sony reports shutter speeds that are integer seconds as "300/10", "250/10", "200/10" etc. This happens both on Sony a6600 I currently have as well as I can see this when searching for 0xd229 among camlibs/sony-*.txt - that is, even models with enums use this format.

Interestigly, setting shutter speeds accept either form, so what happened is you could set shutter speed to e.g. "5" and camera would change it, but then camera would report back "50/10" which is no longer valid value for the shutter speed widget.

This PR normalizes all such shutter speeds to their integer form, so that "5" would be reported back as well in such scenario.


msmeissn:
thx!

PR #689, opened 2021-07-23
Fix memory leak in capture preview for Sony

RReverser:
I've noticed consistent memory increase while streaming preview from a Sony camera. Building with `-fsanitize=leak` revealed that the leak is coming from `strdup <- ptp_unpack_string <- ptp_unpack_OI <- ptp_getobjectinfo`.

Apparently, `PTPObjectInfo` contains heap-allocated strings for `Filename` & `Keywords`, and so `ptp_free_objectinfo` always must be called on the `ptp_getobjectinfo` results.

That all sounds good, but what seriously worries me is that, after fixing this particular codepath for PTP_VENDOR_SONY, I've searched for other usages of `ptp_getobjectinfo`, and majority of them don't ever call `ptp_free_objectinfo` on the result.

That is, there's probably tons of other leaks out there for various cameras just out of this single call, and someone needs to go through all those usages and fix them one by one. While it might be a simple refactoring, it requires great care and I'm not ready to invest into it especially when there's no way of testing the result with other camera models.

Perhaps a more systematic fix is necessary - for example, `PTPObjectInfo` could be changed to contain fixed-length char arrays and sidestep this problem altogether.

More generally, it might be also worth running GPhoto's tests with address & leak sanitizers enabled to detect all such similar problems.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -3603,6 +3603,8 @@ enable_liveview:
 				continue;
 			}
 
+			ptp_free_objectinfo(&oi);
+
 			ret = ptp_getobject_with_size(params, preview_object, &ximage, &size);
 			if (ret == PTP_RC_OK)
 				break;


RReverser:
> mainly the objectinfos go into the PTPObjects and are freed from there.

Maybe I missed a few, but I haven't seem examples of that.

E.g. from the same file, the Fuji switch below: https://github.com/RReverser/libgphoto2/blob/fcdf03182febb8ced418c8a002ca04682b200e86/camlibs/ptp2/library.c#L3692. As far as I can tell, this `oi` will also be leaked like the Sony one.

Same for this one: https://github.com/RReverser/libgphoto2/blob/fcdf03182febb8ced418c8a002ca04682b200e86/camlibs/ptp2/library.c#L4489

Or here: https://github.com/RReverser/libgphoto2/blob/fcdf03182febb8ced418c8a002ca04682b200e86/camlibs/ptp2/library.c#L4893

...and there are many others.

RReverser:
Created a separate tracking issue to continue discussion: https://github.com/gphoto/libgphoto2/issues/690

PR #691, opened 2021-07-26
Add Sony a6600 descriptors

RReverser:
Without this, camera abilities would report a6600 as not having "trigger capture" even though it works on practice.


RReverser:
I'm not entirely sure why this needs a static list though... Can't `gp_camera_get_abilities` read actual camera capabilities from `ptp_operation_issupported`?

msmeissn:
camera_get_abilities does not connect to the camera, it runs without camera connection.

RReverser:
But since it accepts Camera pointer, doesn't it mean camera should be connected by that time anyway?

PR #698, opened 2021-08-05
Fixed build in mingw

yarda:
Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>

### DIFF
--- a/camlibs/lumix/lumix.c
+++ b/camlibs/lumix/lumix.c
@@ -36,10 +36,15 @@
 #include <libxml/xmlreader.h>
 
 
-#include <sys/socket.h>
+#ifdef WIN32
+# include <winsock2.h>
+# include <ws2tcpip.h>
+#else
+# include <sys/socket.h>
+# include <netinet/in.h>
+# include <arpa/inet.h>
+#endif
 #include <stdlib.h>
-#include <netinet/in.h>
-#include <arpa/inet.h>
 
 
 #include <gphoto2/gphoto2-library.h>
--- a/configure.ac
+++ b/configure.ac
@@ -329,6 +329,9 @@ AC_HEADER_DIRENT
 AC_HEADER_STDC
 # after _HEADER_STDC
 AC_CHECK_HEADERS([sys/param.h sys/mman.h sys/select.h locale.h memory.h getopt.h unistd.h mcheck.h limits.h sys/time.h langinfo.h])
+AC_CHECK_LIB(mman, mmap, [
+	AM_CONDITIONAL([HAVE_MMAN_LIB], [test x$ac_cv_lib_mman_mmap = xyes])
+	])
 AC_C_INLINE([])
 AC_C_CONST([])
 dnl FIXME: AC_STRUCT_TIMEZONE
--- a/configure.ac
+++ b/configure.ac
@@ -329,9 +329,19 @@ AC_HEADER_DIRENT
 AC_HEADER_STDC
 # after _HEADER_STDC
 AC_CHECK_HEADERS([sys/param.h sys/mman.h sys/select.h locale.h memory.h getopt.h unistd.h mcheck.h limits.h sys/time.h langinfo.h])
-AC_CHECK_LIB(mman, mmap, [
-	AM_CONDITIONAL([HAVE_MMAN_LIB], [test x$ac_cv_lib_mman_mmap = xyes])
-	])
+
+dnl If sys/mman.h is present, check whether mmap requires the mman
+dnl library (see camlibs/st2205/st2205.c st2205_malloc_page_aligned()).
+MMAP_LIBS=""
+AS_VAR_IF([ac_cv_header_sys_mman_h], [yes], [dnl
+gp_saved_LIBS="$LIBS"
+LIBS=""
+AC_SEARCH_LIBS([mmap], [mman])
+MMAP_LIBS="$LIBS"
+LIBS="$gp_saved_LIBS"
+])dnl
+AC_SUBST([MMAP_LIBS])
+
 AC_C_INLINE([])
 AC_C_CONST([])
 dnl FIXME: AC_STRUCT_TIMEZONE
--- a/libgphoto2/gphoto2-setting.c
+++ b/libgphoto2/gphoto2-setting.c
@@ -37,7 +37,7 @@
 #ifdef WIN32
 /* Win32 headers may use interface as a define; temporarily correct this */
 #define interface struct
-#include <Shlobj.h>
+#include <shlobj.h>
 #undef interface
 #endif
 
--- a/camlibs/st2205/Makefile-files
+++ b/camlibs/st2205/Makefile-files
@@ -10,3 +10,7 @@ st2205_la_LDFLAGS = $(camlib_ldflags)
 st2205_la_DEPENDENCIES = $(camlib_dependencies)
 st2205_la_LIBADD = $(camlib_libadd) @LIBGD_LIBS@ $(LTLIBICONV)
 st2205_la_CFLAGS = @LIBGD_CFLAGS@
+
+if HAVE_MMAN_LIB
+st2205_la_LIBADD += -lmman
+endif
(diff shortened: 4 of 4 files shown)


camlibs/Makefile.am
```diff
@@ -68,6 +68,9 @@ camlib_ldflags += -avoid-version
 camlib_ldflags += -export-dynamic
 camlib_ldflags += -export-symbols '$(top_srcdir)/camlibs/camlib.sym'
 camlib_ldflags += -rpath '$(camlibdir)'
+if HAVE_MMAN
```
yarda, line 71: On mingw it needs linking with the libmman for the mmap/munmap which are needed by the st2205 camlib driver. This is simple workaround how to do it.

camlibs/lumix/lumix.c
```diff
@@ -36,10 +36,15 @@
 #include <libxml/xmlreader.h>
 
 
-#include <sys/socket.h>
```
yarda, line 39: The lumix driver needs ifdef for the socket includes like the other drivers use. I don't have lumix, thus I couldn't check the driver, but it now compiles at least.

configure.ac
```diff
@@ -329,6 +329,7 @@ AC_HEADER_DIRENT
 AC_HEADER_STDC
 # after _HEADER_STDC
 AC_CHECK_HEADERS([sys/param.h sys/mman.h sys/select.h locale.h memory.h getopt.h unistd.h mcheck.h limits.h sys/time.h langinfo.h])
+AM_CONDITIONAL([HAVE_MMAN], [test x$ac_cv_header_sys_mman_h = xyes])
```
yarda, line 332: The AM define for the camlibs/Makefile.am which is set according to the sys/mman.h header detection. There are better ways how to do it, but this one is simple and worked for me.

libgphoto2/gphoto2-setting.c
```diff
@@ -37,7 +37,7 @@
 #ifdef WIN32
 /* Win32 headers may use interface as a define; temporarily correct this */
 #define interface struct
-#include <Shlobj.h>
```
yarda, line 40: On Windows the case doesn't matter, but it is not the case on Linux. It seems the mingw provides the header in the lowercase (at least on Fedora 33), so use the lowercase.

ndim (commented):
I presume you ran your build tests without a `--with-camlibs=` argument to `configure` or with `--with-camlibs=standard`?

We should at least out of interest check whether `--with-camlibs=everything` works here or not, even if we do not port/fix-to-compile all the obsolete drivers.

However, on my Fedora 34 system, even `mingw32-configure && mingw32-make` fails very quickly due to conflicting types for `ssize_t`, and `mingw64-configure && mingw64-make` fails a little bit later when compiling `libgphoto2_port/usb/libusb.c`.

How are you running your test builds?

yarda:
> I presume you ran your build tests without a `--with-camlibs=` argument to `configure` or with `--with-camlibs=standard`?
> 
I tried both, both worked, I presume the missing argument defaults to `--with-camlibs=standard`?

> We should at least out of interest check whether `--with-camlibs=everything` works here or not, even if we do not port/fix-to-compile all the obsolete drivers.
> 
I didn't try this, I will check.

> However, on my Fedora 34 system, even `mingw32-configure && mingw32-make` fails very quickly due to conflicting types for `ssize_t`, and `mingw64-configure && mingw64-make` fails a little bit later when compiling `libgphoto2_port/usb/libusb.c`.
> 
> How are you running your test builds?

I built using mxe, with the patch it builds fine for win64. I created upstream PR (I haven't tested yet the functionality on the win64):
https://github.com/mxe/mxe/pull/2697

camlibs/Makefile.am
```diff
@@ -68,6 +68,9 @@ camlib_ldflags += -avoid-version
...
+camlib_ldflags += -lmman
```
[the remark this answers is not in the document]
yarda: OK, it was just quick and dirty fix. I will update the PR.

yarda:
I can confirm with `--with-camlibs=everything` it compiled OK.

ndim:
> > I presume you ran your build tests without a `--with-camlibs=` argument to `configure` or with `--with-camlibs=standard`?
> 
> I tried both, both worked, I presume the missing argument defaults to `--with-camlibs=standard`?

That is correct.

> > We should at least out of interest check whether `--with-camlibs=everything` works here or not, even if we do not port/fix-to-compile all the obsolete drivers.

> I can confirm with `--with-camlibs=everything` it compiled OK.

Thank you very much!

camlibs/Makefile.am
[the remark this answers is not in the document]
ndim: I have just pushed the fix.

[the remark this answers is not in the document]
yarda: Thanks, you saved me some time. I can confirm it still builds OK with the 'everything'. I will update the mxe PR.

yarda:
It seems the CI build now fails with the:
```
mv -f st2205/.deps/la-st2205_tables.Tpo st2205/.deps/la-st2205_tables.Plo
/usr/bin/ld: cannot find -lmman
```

yarda:
> It seems the CI build now fails with the:
> 
> ```
> mv -f st2205/.deps/la-st2205_tables.Tpo st2205/.deps/la-st2205_tables.Plo
> /usr/bin/ld: cannot find -lmman
> ```

Got it, it seems mman.h exists on the linux, so the current check with the -lmman is not good. We probably need to check for the win.

yarda:
I squashed the commits into one and tried to fix the CI error.

configure.ac
ndim: On my Fedora 34 system, running `../configure --with-camlibs=st2205` for a native build of only the camlib in question prints the following:

```
checking sys/mman.h usability... yes
checking sys/mman.h presence... yes
checking for sys/mman.h... yes
```

and then the `make` run fails with

```
/usr/bin/ld: cannot find -lmman
```

as there is no `libmman` on the system to link `st2205.so` against.

This definition of the `HAVE_MMAN` automake conditional may be simple, but it does *not* work for me for native Linux builds. The check should be whether linking a piece of code which calls `mmap(2)` works without adding `-lmman` to the libraries to link against, and whether linking against `-lmman` actually fixes that. And such a condtional would be more accurately named something like `MMAP_NEEDS_LIBMMAN`.

However, we should take a step back here or two to see what the `st2205` camlib actually does with `mmap(2)`: It allocates a page aligned memory buffer by opening `/dev/zero`. Does that trick even work on Windows?

The comment above the `st2205_malloc_page_aligned(int size)` function (yes, `int` not `size_t`) indicates that the page aligned buffer is for cooperating with the use of `O_DIRECT` within `usbdiskdirect` iolib. This iolib is implemented by a source file `libgphoto2_port/usbdiskdirect/linux.c` whose name suggests it has been written very much with Linux in mind, and which just defines the `O_DIRECT` macro to be `00040000` if it is undefined. What that value does on Windows needs to be tested. https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-160 indicates that `O_DIRECT` does not exist on Windows.

So I would like to ask whether it even makes sense at all to just link `st2205.so` against `libmman` in a mingw build (`--with-camlibs=standard,-st2205`), or whether the `st2205` driver requires so much porting effort that it should be just deactivated in mingw builds.

yarda: The problem with the -lmman breaking the native build should be addressed with the latest PR. But of course I can drop the st2205 driver in the mxe (that's what I initially did).

libgphoto2/gphoto2-setting.c
ndim: Looks obvious.

camlibs/lumix/lumix.c
ndim: Appears sensible.

ndim:
I would be fine with the code as of my commit bbac625 I have pushed. Does that build for you with MXE?

yarda:
> I would be fine with the code as of my commit [bbac625](https://github.com/gphoto/libgphoto2/commit/bbac625a7ea74decb552bf0b497ed65374b62e88) I have pushed. Does that build for you with MXE?

Yes, it builds OK, thanks.

PR #699, opened 2021-08-06
camlibs/ptp2/ptp-pack.c: Fix aperture on EOS RP

jktjkt:
My wife's EOS RP (FW 1.4.0) always reports the lens' lowest aperture unless we apply the same special handling which is enabled for `olcver == 0x13` (apparently, EOS R5). Tested with RF 35/1.8, and also on 5Div + Sigma 35 Art (to check a possible regression).

This reverts commit 5bf98e7f09e82d0e27f41a8df4ccb10e5feb6944 (and fixes a comment so that it more accurately reflects what's going on). I'm a bit nervous about this because that commit looks like it was explicitly designed to do the exact opposite of this one, but perhaps it was just to be extra cautious?

Cc: @arkellr @msmeissn


msmeissn:
i think it was overly cautios. appliemd thanks

PR #701, opened 2021-08-06
gh workflows: remove requirements for flex, bison

ndim:
As 4316f75ed4ce7a4ae5296264699665631c4e62d6 has removed gp2ddb with its use of `flex` and `bison`, we do not need `flex` and `bison` any more, and so the github workflows do not need to install `flex` and `bison` either.

I have made this a PR mainly to get github to actually run the workflows to check.


msmeissn:
lgtm if you want to merge go ahead

Issue #251, opened 2018-03-25
Windows Port - CAMLIBS Default Path Issue


Oxydation:
I am currently working on a project to provide `libgphoto2 `on multiple platforms via C#. To achieve this I am working with [SharpCamera](https://gitlab.com/define-private-public/SharpCamera) from @define-private-public. Despite some startup problems I got it running from C# code on a Win 64-bit machine with 32-bit dlls of `libgphoto2 `(check this [issue](https://gitlab.com/define-private-public/SharpCamera/issues/10)). 

Future goal is to ship a NuGet package with all dependencies. We want to make sure, that there is less or no setup hassle for users. I succeeded in putting all dependencies (`libgphoto2-6.dll, libgphoto2_port-12.dll, libexif-12.dll, libintl-8.dll, libsystre-0.dll, libgcc_s_dw-1.dll`) as well as the IOLIBS (`usb1.dll`, `disk.dll` and `ptpip.dll`) right next to the application. I am using those libraries provided by the MYSYS2, MINGW32 environment.

Now to my problem: I could not load the camlibs if I put them next to the` libgphoto2-6.dll` or in a `CAMLIBS `directory in the same directory of `libgphoto2-6.dll`. Please check the given [issue](https://gitlab.com/define-private-public/SharpCamera/issues/10#note_64959496) for more information. 

It seems that during the build process a not existing path has been added to the libgphoto2-6.dll, where instead should be just a dot `.`. It represents the value of the `CAMLIBS ` define. As you can see in the issue, I had to modify the DLL via hex editor to change the wrong path to the` .` (dot). After this modification I can run the application with all cam-libs relative to the `libgphoto2-6.dll`.
![image](https://user-images.githubusercontent.com/8331862/37878536-0dbbad5e-306b-11e8-80cf-5dd055ae52e6.png)

Some - maybe important - detail: I can set the environment variable in Windows (CAMLIBS, `.`) then it was also working. But if the environment variable was not set, it used the weird path from the DLL.

Oxydation:
Yes, I was able to use SharpCamera on Windows, as stated in the issue and the linked issue of the SharpCamera-Repo.

define-private-public:
@Oxydation What are the changes you've made?  I'd really like to get SharpCamera to work properly on Windows.

Oxydation:
All changes done are in my fork of SharpCamera under the branch **wintest**: https://gitlab.com/Oxydation/SharpCamera/tree/wintest

[...]

PR #705, opened 2021-08-11
On win change default camlibs/iolibs directories

yarda:
Currently, when cross-compiling for the win, it installs camlibs/iolibs to the cross-compiler environment, e.g. for the camlibs some strange path like `/home/user/crosscompiler-env/.../libgphoto2/2.5.27/`. But later when the bundled resulting package is installed on the win, the path needs to something like `C:\Program Files\Program\`. For the win it seems libgphoto2 defines the default camlibs/iolibs load dir as `.`, which should work. Unfortunately, when bundling packages for win, there can be hundreds of DLLs, which may slow down startup of the libgphoto2, because it's testing the DLLs, whether they are camlibs/iolibs. Defining the environment variables IOLIBS/CAMLIBS_ENV works, but it's complication.

This PR propose changing the default runtime load dir to the `./libgphoto2` for camlibs and `./libgphoto2_port` for iolibs.

But there is another problem, because the `configure` script doesn't have switch for the runtime dir and overrides the code defaults with the installdir. I.e. it explicitly adds to the CFLAGS `-DIOLIBS=IOLIBS_INSTALL_PATH` and `-DCAMLIBS=CAMLIBS_INSTALL_PATH` that are the 'strange' paths to the cross compiler. I temporally resolved it by:
```
--- a/configure.ac
+++ b/configure.ac
@@ -521,7 +534,6 @@ AC_ARG_WITH([camlibdir],[AS_HELP_STRING(
 ])
 AC_MSG_RESULT([${camlibdir}])
 AC_SUBST([camlibdir])
-AM_CPPFLAGS="$AM_CPPFLAGS -DCAMLIBS=\\\"\$(camlibdir)\\\""
 


--- a/libgphoto2_port/configure.ac
+++ b/libgphoto2_port/configure.ac
@@ -468,7 +468,6 @@ for x in ${IOLIB_LIST}; do
 done
 AC_SUBST(IOLIB_LTLIST)
 AC_SUBST([iolibdir],["\$(libdir)/\$(PACKAGE_TARNAME)/\$(VERSION)"])
-AM_CPPFLAGS="$AM_CPPFLAGS -DIOLIBS=\\\"${iolibdir}\\\""
 
 sorted_iolib_list="$(echo "${IOLIB_LIST}" | tr ' ' '\n' | sort | ${SED} '/^$/d' | tr '\n' ' ' | ${SED} 's/ $//')"
 AC_DEFINE_UNQUOTED([IOLIB_LIST], ["${sorted_iolib_list}"], [Define as string containing a list of the iolibs])
```

I think this should be resolved somehow, e.g.:
- on win do not explicitly add the -DIOLIBS/CAMLIBS
- add `configure` switch allowing override of the runtime load dirs

There is e.g. #251 which is related.

### DIFF
--- a/gphoto2/gphoto2-abilities-list.h
+++ b/gphoto2/gphoto2-abilities-list.h
@@ -196,6 +196,11 @@ const char *gp_message_codeset (const char *);
  */
 #ifdef _GPHOTO2_INTERNAL_CODE
 #define CAMLIBDIR_ENV "CAMLIBS"
+#ifdef WIN32
+#ifndef CAMLIBS
+#define CAMLIBS "./libgphoto2"
+#endif
+#endif
 #endif /* _GPHOTO2_INTERNAL_CODE */
 
 
--- a/gphoto2/gphoto2.h
+++ b/gphoto2/gphoto2.h
@@ -39,7 +39,7 @@ extern "C" {
 
 #ifdef WIN32
 #ifndef CAMLIBS
-#define CAMLIBS "."
+#define CAMLIBS "./libgphoto2"
 #endif
 #endif
 
--- a/libgphoto2_port/gphoto2/gphoto2-port-portability.h
+++ b/libgphoto2_port/gphoto2/gphoto2-port-portability.h
@@ -38,7 +38,7 @@
 # include <direct.h>
 
 # ifndef IOLIBS
-#  define IOLIBS			"."
+#  define IOLIBS		"./libgphoto2_port"
 # endif
 # define strcasecmp		_stricmp
 # ifndef snprintf


gphoto2/gphoto2-abilities-list.h
```diff
@@ -196,6 +196,11 @@ const char *gp_message_codeset (const char *);
  */
 #ifdef _GPHOTO2_INTERNAL_CODE
 #define CAMLIBDIR_ENV "CAMLIBS"
+#ifdef WIN32
```
yarda, line 199: This should be defined here, because it's only defined by the -DCAMLIBS CFLAG. Maybe, it could include the `gphoto2.h`?

msmeissn:
dont really know either way if this breaks things... hmm

yarda:
> dont really know either way if this breaks things... hmm

It probably could break things.

What about the `gphoto2/gphoto2-abilities-list.h`? I think it should define the CAMLIBS or include `gphoto2.h`.

Also I think possible solution could be adding `configure` option for compiling in arbitrary runtime directories.

yarda:
> > dont really know either way if this breaks things... hmm
> 
> It probably could break things.
> 
> What about the `gphoto2/gphoto2-abilities-list.h`? I think it should define the CAMLIBS or include `gphoto2.h`.
> 
> Also I think possible solution could be adding `configure` option for compiling in arbitrary runtime directories.

I.e. to have possibility to have different install dir and runtime dir, which now isn't possible.

ndim:
Uhmm. And why doesn't setting the camlib dir and iolib dir to your desired values "./libgphoto2" and "./libgphoto2_port" work?

```
configure camlibdir="./libgphoto2" iolibdir="./libgphoto2_port"
```

yarda:
> Uhmm. And why doesn't setting the camlib dir and iolib dir to your desired values "./libgphoto2" and "./libgphoto2_port" work?
> 
> ```
> configure camlibdir="./libgphoto2" iolibdir="./libgphoto2_port"
> ```

I am not sure where it will install it, but it shouldn't be problem, I can change the win bundling scripts to grab it from arbitrary location. I will check this, it could work for me.

ndim:
We could add `CAMLIBS` and `IOLIBS` as `configure` variables, defaulting to `$(camlibdir)` and `$(iolibdir)`, but settable at configure time. The values could then be defined as `-DCAMLIBS=...` and `-DIOLIBS=` macros.

yarda:
> > Uhmm. And why doesn't setting the camlib dir and iolib dir to your desired values "./libgphoto2" and "./libgphoto2_port" work?
> > ```
> > configure camlibdir="./libgphoto2" iolibdir="./libgphoto2_port"
> > ```
> 
> I am not sure where it will install it, but it shouldn't be problem, I can change the win bundling scripts to grab it from arbitrary location. I will check this, it could work for me.

It worked for me, I will PR this change to mxe.

yarda:
> We could add `CAMLIBS` and `IOLIBS` as `configure` variables, defaulting to `$(camlibdir)` and `$(iolibdir)`, but settable at configure time. The values could then be defined as `-DCAMLIBS=...` and `-DIOLIBS=` macros.
> 
> Then you could statically set the install directory and the runtime lookup location at `configure` time:
> 
> ```
> configure CAMLIBS=./libgphoto2 camlibdir=/path/to/wherever
> ```

I think this would be nice improvement, definitely +1 from me.

yarda:
Now, what to do with the `gphoto2/gphoto2-abilities-list.h`? I think it should be consistent with the other code. I.e. if the other code defines CAMLIBS for win to be '.' if undefined (e.g. here https://github.com/gphoto/libgphoto2/blob/31f30208ec1af1baac6bc6535034f173f9bd21ef/gphoto2/gphoto2.h#L42), I think it should do the same and not rely on the -DCAMLIBS. I can change this PR this way to fix this or close it.

yarda:
> > > Uhmm. And why doesn't setting the camlib dir and iolib dir to your desired values "./libgphoto2" and "./libgphoto2_port" work?
> > > ```
> > > configure camlibdir="./libgphoto2" iolibdir="./libgphoto2_port"
> > > ```
> > 
> > 
> > I am not sure where it will install it, but it shouldn't be problem, I can change the win bundling scripts to grab it from arbitrary location. I will check this, it could work for me.
> 
> It worked for me, I will PR this change to mxe.

I was too fast, it didn't override the path.

ndim:
We now allow setting the default runtime lookup location `DEFAULT_(CAMLIBS|IOLIBS)` independently from the installation location `(camlib|iolib)dir` at configure time:

    ./configure DEFAULT_CAMLIBS=./whatever

or

    ./configure DEFAULT_CAMLIBS=./whatever camlibdir=/path/to/camlibs

Does that solve your issue?

yarda:
I think it's OK, thanks.

PR #707, opened 2021-08-12
Add add missing ISO values from Canon EOS R6

trilader:
The Canon EOS R6 has two additional ISO values that were not supported before: 64000 and 80000.


msmeissn:
thanks!

PR #725, opened 2021-10-05
change all files to the utf-8 text encoding

ndim:
This changes all files in the source tree to the utf-8 text encoding. Currently, the source tree contains a mix of UTF-8 and ISO-8859-1 files, as you can check with `file`, `git`, and `grep`:

```
[user@host libgphoto2]$ file $(git ls-files) | grep 'UTF'
[user@host libgphoto2]$ file $(git ls-files) | grep 'ISO'
```

This converts the remaining ISO-8859-x files to UTF-8.


ndim:
Out of curiosity in connection with https://github.com/gphoto/libgphoto2/issues/733, I have taken a look at the Translation Project's status pages for libgphoto2, libgphoto2_port, and gphoto2, and they all list the German `de` translations as "external", i.e. someone *outside* the Translation Project handles the `de.po` files, i.e. there probably are no external submissions of `de.po` files which we just copy into our source tree verbatim.

  * https://translationproject.org/domain/libgphoto2.html
  * https://translationproject.org/domain/libgphoto2_port.html
  * https://translationproject.org/domain/gphoto2.html

Judging from the respective `de.po` files' headers, @msmeissn was the last translator for gphoto2 and libgphoto2, and from the copyright notices for libgphoto2_port as well.

Of these three, only `libgphoto2_port/po/de.po` is actually still written in ISO-8859-1, so I will change that to UTF-8 shortly.

This will leave `libgphoto2_port/po/eu.po` (Basque, last translated for 2.5.27) as the only remaining ISO-8859 po file inside the libgphoto2 repo, and three languages (es fi sk Spanish, Finnish, Slovak, last translated for 2.5.9) for the gphoto2 CLI.

PR #731, opened 2021-10-12
Merge libgphoto2_port buildsystem into the libgphoto2 buildsystem

ndim:
This will merge the `libgphoto2_port/configure.ac` and assorted files into the top-level `configure.ac`. The basic idea is to get both `configure.ac` and `libgphoto2_port/configure.ac` cleaned up first, both to refamiliarize myself with them and to make the later merge simpler. Only then will I do the actual merge.

**At this time, this is still work in progress. There will be force pushes for some time.**

Changes and Features:
  * `configure --help` will now print the help for the `libgphoto2_part` in addition to the `libgphoto2` part of the build.
  * Running `configure` should be faster.
  * Require newer autoconf version. Old requirement was 2.62 (from 2008-04-06). New requirement is 2.69 (from 2012-04-25, like libexif).
  * Require newer automake version. Old requirement was 1.9 (from 2004-07-28). New requirement is 1.14 (from 2013-06-20, as opposed to libexif requiring 1.14.1).
  * Require newer gettext version. Old requirement was 0.14.1 (from 2004-01-30), new requirement will be 0.18.3 (from 2013-07-07, like libexif) or 0.19.1 (from 2014-06-10, allows `make dist` without updating all the `*.po` files). Exact version TBD.
  * Will require newer libtool (TBD).
  * Removed some hacks required to work with tools from early 2000s Debian oldoldstable.

Remarks:
  * As of 2021-10-15, Debian oldoldstable ships the following tool versions: autoconf 2.69, automake 1.15, gettext 0.19.8.1, and libtool 2.4.6. Being quite old by definition, Debian oldoldstable should not be a bad place to take a hint about tool versions from.
  * Also regarding build tool versions, you only need those tools to build a dist tarball from a git working tree. That release tarball (or any of the official release tarballs) should then build on even older systems.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -225,15 +225,14 @@ GP_VA_COPY
 
 
 dnl check for/set up libtool and libltdl
-dnl AC_DISABLE_STATIC
-dnl AC_DISABLE_SHARED
-dnl AC_LIBLTDL_CONVENIENCE([libltdl])
-dnl AC_WITH_LTDL
-dnl AC_LIB_LTDL
-AC_DISABLE_STATIC 
-AC_LIBTOOL_WIN32_DLL
-AC_LIBTOOL_DLOPEN
-AC_PROG_LIBTOOL
+LT_PREREQ([2.4.2])
+dnl Disable building static library, as no one uses it anyway.
+LT_INIT([
+  disable-static
+  dlopen
+  win32-dll
+])
+dnl LTDL_INIT([external])
 
 dnl We are using our own libltdl checks instead of AC_WITH_LTDL
 dnl because we do not want to ship our own copy of libltdl.
--- a/libgphoto2_port/configure.ac
+++ b/libgphoto2_port/configure.ac
@@ -78,16 +78,14 @@ AM_PROG_AR
 
 
 dnl check for/set up libtool and libltdl
-dnl AC_DISABLE_STATIC
-dnl AC_DISABLE_SHARED
-dnl AC_LIBLTDL_CONVENIENCE([libltdl])
-dnl AC_WITH_LTDL
-dnl AC_LIB_LTDL
+LT_PREREQ([2.4.2])
 dnl Disable building static library, as no one uses it anyway.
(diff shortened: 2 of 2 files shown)


ndim:
As of cf50ed6, this needs probably another 20 to 30 commits until the actual merge, and then another 10 to clean up the merge fallout. Still working on this in the background, occasionally committing smaller independent issues I have run into directly to the master branch.

ndim:
Actually unifying the two `configure.ac` scripts into one will *NOT* happen in time for 2.5.28, but many of the preparations will.

ndim:
As PR #855 shows, unifying the two `configure.ac` scripts is also useful to eliminate the need to duplicate code and the chance to build in inconsistent environments.

ndim:
FTR, I am picking this up again.

axxel:
Random thought from an outsider: maybe better replace the whole thing with e.g. CMake? That is at least my thought every time I look at my console while tons of 'noise' gets spit out by make and I have to scroll just to find some warnings... ;).

Disclaimer: I have no idea how much work that would be, I'm just a happy CMake user in other projects.

hfiguiere:
>  maybe better replace the whole thing with e.g. CMake?

HELL NO

axxel:
> HELL NO

Ah, someone who likes their autotools ;). Maybe you can help me solve my problem regarding the usability. Is there an autoconf/configure/make option to do away will all the noise and have a make run with a single compiler warning actually let me see that in my terminal window without scrolling 10 to 30 pages up?

msmeissn:
make --quiet

already helps.

axxel:
Thanks for the hint, although it made no significant difference IMHO. But now I found `configure --enable-silent-rules` and that basically does what I wanted to achieve with my cmake suggestion.

hfiguiere:
this should be enabled by default.

ndim:
Enable silent rules by default PR: https://github.com/gphoto/libgphoto2/pull/1021

PR #740, opened 2021-11-05
Fix __VA_ARGS__ in ptp2 for Android NDK compiler

fahrion2600:
The Android NDK compiler treats `__VA_ARGS__` differently than gcc when passing
empty arguments. In particular, `NARGS()` returns `1` instead of `0`. This causes an
off-by-one bug in the `ptp2` camlib driver.

I'm not an expert in C macros, so I took the easy/ugly path of splitting the
`PTP_CNT_INIT` macro into two, the original for no params and `PTP_CNT_INIT_ARGS`
for one or more params.

Tested:
* Compiled successfully
* From my rooted Android phone, I tested triggering captures from USB and WiFi
using PTP/IP.

### DIFF
--- a/camlibs/ptp2/ptp.c
+++ b/camlibs/ptp2/ptp.c
@@ -75,10 +75,16 @@ ptp_init_container(PTPContainer* ptp, uint16_t code, int n_param, ...)
 	va_end(args);
 }
 
+// Android NDK's compiler treats no-argument __VA_ARGS__ differently, such that
+// NARGS() returns 1 instead of 0. To avoid this, separate PTP_CNT_INIT into
+// no-argument and variadic-argument versions.
+#define PTP_CNT_INIT(PTP, CODE) \
+	ptp_init_container(&PTP, CODE, 0)
+
 #define NARGS_SEQ(_1,_2,_3,_4,_5,_6,_7,_8,N,...) N
 #define NARGS(...) NARGS_SEQ(-1, ##__VA_ARGS__, 7, 6, 5, 4, 3, 2, 1, 0)
 
-#define PTP_CNT_INIT(PTP, CODE, ...) \
+#define PTP_CNT_INIT_ARGS(PTP, CODE, ...) \
 	ptp_init_container(&PTP, CODE, NARGS(__VA_ARGS__), ##__VA_ARGS__)
 
 static uint16_t ptp_exit_recv_memory_handler (PTPDataHandler*,unsigned char**,unsigned long*);
@@ -503,7 +509,7 @@ ptp_getstreaminfo (PTPParams *params, uint32_t streamid, PTPStreamInfo *si)
 	unsigned int	size;
 	int		ret;
 
-	PTP_CNT_INIT(ptp, PTP_OC_GetStreamInfo, streamid);
+	PTP_CNT_INIT_ARGS(ptp, PTP_OC_GetStreamInfo, streamid);
 	CHECK_PTP_RC(ptp_transaction(params, &ptp, PTP_DP_GETDATA, 0, &data, &size));
 	ret = ptp_unpack_StreamInfo(params, data, si, size);
 	free (data);
@@ -540,7 +546,7 @@ ptp_canon_eos_905f (PTPParams* params, uint32_t x)
 	unsigned char	*data = NULL;
 	unsigned int	size;
 
-	PTP_CNT_INIT(ptp, 0x905f, x);
+	PTP_CNT_INIT_ARGS(ptp, 0x905f, x);
 	CHECK_PTP_RC(ptp_transaction(params, &ptp, PTP_DP_GETDATA, 0, &data, &size));
 	free (data);
 	return PTP_RC_OK;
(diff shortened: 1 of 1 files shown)


fahrion2600:
I'm open to suggestions on making this cleaner.

ndim:
I have had a go at this and changed the macro and function definition from

    #define PTP_CNT_INIT(CNT, CODE, ...)
    void ptp_init_container(PTPContainer* ptp, uint16_t code, int n_param, ...)

to 

    #define PTP_CNT_INIT(CNT, ...)
    void ptp_init_container(PTPContainer* ptp, int n_param, ...)

and adapted the inside of `ptp_init_container()` as well. This means that there is no need to have one macro for the no argument case and another for the one-or-more arguments case.

The macro test program looks like

```
#include <assert.h>
#include <stdarg.h>
#include <stdio.h>


// #define TRY_CASES_OUTSIDE_VALID_RANGE


#define N_PARAM_SEQ(DUMMY,				\
		    Pa, Pb, Pc, Pd, Pe,			\
		    P0, P1, P2, P3, P4, P5,		\
		    N, ...)				\
  N


#define N_PARAM(...)				\
  N_PARAM_SEQ(__VA_ARGS__,			\
	      "a", "b", "c", "d", "e",		\
	      5, 4, 3, 2, 1, 0, NULL)


#define FOO_INIT(PTR, ...) \
  foo_func(PTR, N_PARAM(-666, __VA_ARGS__), __VA_ARGS__)


static
void foo_func(void *ptr, int n_param, ...)
{
  va_list args;

  (void) ptr;

  printf("%s: n_param=%d\n", "foo_func", n_param);
  assert(n_param <= 5);

  va_start(args, n_param);
  int code = va_arg(args, int);
  printf("  code   = %d\n", code);

  for (int i=0; i<n_param; ++i) {
    int value = va_arg(args, int);
    printf("  Param%d = param[%d] = %d\n", i+1, i, value);
  }
  va_end(args);
}


int main()
{
  /* zero to five parameters after the code are the valid use cases */
  FOO_INIT(NULL, 1000);
  FOO_INIT(NULL, 1001, 201);
  FOO_INIT(NULL, 1002, 201, 202);
  FOO_INIT(NULL, 1003, 201, 202, 203);
  FOO_INIT(NULL, 1004, 201, 202, 203, 204);
  FOO_INIT(NULL, 1005, 201, 202, 203, 204, 205);

  /* from here on, we want compile time errors or at least runtime
     errors */
#ifdef TRY_CASES_OUTSIDE_VALID_RANGE
  FOO_INIT(NULL, 1006, 201, 202, 203, 204, 205, 206);
  FOO_INIT(NULL, 1007, 201, 202, 203, 204, 205, 206, 207);
  FOO_INIT(NULL, 1008, 201, 202, 203, 204, 205, 206, 207, 208);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212, 213);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212, 213, 214);
  FOO_INIT(NULL, 1009, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212, 213, 214, 215);
#endif

  return 0;
}
```

I have tested this test program using `clang` and `gcc` on my glibc PC Linux system (most PC Linux distros are using gcc these days, but Android NDK is using clang), and I hope removing the `##` part from `##__VA_ARGS__` removed the different argument counts:

  * `clang -std=c99 -pedantic -Wall -Wextra -Weverything -o n-param-test-clang n-param-test.c && ./n-param-test-clang`
  * `gcc -std=c99 -pedantic -Wall -Wextra -o n-param-test-gcc n-param-test.c && ./n-param-test-gcc`

Both compilers produce a program which produces the same output, and I hope that the program compiling without *any* warnings from `clang` means that it should work on PC-clang, PC-gcc, NDK-clang, and hopefully every other compiler as well.

```
foo_func: n_param=0
  code   = 1000
foo_func: n_param=1
  code   = 1001
  Param1 = param[0] = 201
foo_func: n_param=2
  code   = 1002
  Param1 = param[0] = 201
  Param2 = param[1] = 202
foo_func: n_param=3
  code   = 1003
  Param1 = param[0] = 201
  Param2 = param[1] = 202
  Param3 = param[2] = 203
foo_func: n_param=4
  code   = 1004
  Param1 = param[0] = 201
  Param2 = param[1] = 202
  Param3 = param[2] = 203
  Param4 = param[3] = 204
foo_func: n_param=5
  code   = 1005
  Param1 = param[0] = 201
  Param2 = param[1] = 202
  Param3 = param[2] = 203
  Param4 = param[3] = 204
  Param5 = param[4] = 205
```

Could you please verify that this test program compiles and runs properly with Android NDK? That would be awesome as I do not have a working Android NDK installed right now.

fahrion2600:
Hi @ndim,

Thanks for looking into this! That's a very elegant solution! I compiled it with NDK and ran it on my phone, and here's the output:

```
$ /data/local/tmp/n-param-test                                                                                                                                                                  
foo_func: n_param=0
  code   = 1000
foo_func: n_param=1
  code   = 1001
  Param1 = param[0] = 201
foo_func: n_param=2
  code   = 1002
  Param1 = param[0] = 201
  Param2 = param[1] = 202
foo_func: n_param=3
  code   = 1003
  Param1 = param[0] = 201
  Param2 = param[1] = 202
  Param3 = param[2] = 203
foo_func: n_param=4
  code   = 1004
  Param1 = param[0] = 201
  Param2 = param[1] = 202
  Param3 = param[2] = 203
  Param4 = param[3] = 204
foo_func: n_param=5
  code   = 1005
  Param1 = param[0] = 201
  Param2 = param[1] = 202
  Param3 = param[2] = 203
  Param4 = param[3] = 204
  Param5 = param[4] = 205
```

I diff'ed it against your output, and it looks like it matches, so I think your approach works!

ndim:
Thanks for testing, for proposing this PR and for your general efforts in porting to Android.

I plan on polishing my approach a bit more and then push it in a day or two, closing this PR.

For future reference: This problem would not have been an issue at all if the code had been compiling with zero warnings.

msmeissn:
i did not have any warnings at least on this... perhaps some warning combination was missing. thanks for fixing it!

ndim:
Using `##` in front of of `__VA_ARGS__` is a gccism which both gcc and clang warn about when you compile with `-std=c99 -pedantic -Wall -Wextra` (with an additional `-Weverything` thrown in for clang).

PR #743, opened 2021-11-11
print-camera-list: clarify help

jtojnar:
Use “version” instead of “mode” to avoid confusion with the “(file) mode” argument.


ndim (approved):
Looks good to me.

PR #750, opened 2021-12-09
Add Nikon Zfc support

odlg:
Please add Nikon Zfc support. With this applied the Zfc no longer reports 

```
*** Error ***              
liveviewsize not found in configuration tree.
```
when doing a 
`gphoto2 --set-config liveviewsize=2 --stdout --capture-movie `


msmeissn:
added it, thqanks!

PR #758, opened 2022-01-06
Fix lumix linking on mingw

kmilos:
This is to complement https://github.com/gphoto/libgphoto2/commit/5c8c0538714b170ad83b2095dcf42838d663a3d3


kmilos:
Please also consider upstreaming some other mingw related patches:

https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-libgphoto2
https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-gphoto2

msmeissn:
i merged some easier ones. 
but some are hacks that hide other bugs... not sure if there are any linkbacks to why it was fixed there?

kmilos:
The lumix one has indeed been addressed by 5c8c053 already.

The popt patch for gphoto2 is also gone in 2.5.28.

Which leaves only a couple, and I have no reference as to why they were required, sorry.

Issue #717, opened 2021-09-09
libgphoto2+gphoto2 builds fail looking for gphoto2/gphoto2-foo.h include file


ndim:
**Describe the bug**

When there are no `gphoto2/gphoto2-*.h` include files in the C preprocessor's default include path, building gphoto2 fails when trying to `#include` a `gphoto2/gphoto2-foo.h` include file:

```
In file included from actions.c:45:
./actions.h:24:10: fatal error: 'gphoto2/gphoto2-camera.h' file not found
#include <gphoto2/gphoto2-camera.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

The reason is that both `libgphoto2.pc` and `libgphoto2_port.pc` both contain

    Cflags: -I${includedir}/gphoto2

but the gphoto2 source code uses

    #include <gphoto2/gphoto2-camera.h>

which we have declared the standard way of including libgphoto2 headers.

We have removed the symlink which used to link `${includedir}/gphoto2/gphoto2` to `${includedir}/gphoto2` in https://github.com/gphoto/libgphoto2/pull/452 after https://github.com/gphoto/libgphoto2/issues/444, and that symlink used to hack this to work.

**libgphoto2 and gphoto2 version**

**To Reproduce**

On a system which does not have any gphoto2 headers installed in the C preprocessor path,

```
cd
cd libgphoto2
./configure --prefix=$HOME/.local
make all
make install

cd
cd gphoto2
./configure --prefix=$HOME/.local PKG_CONFIG_PATH=$HOME/.local/lib/pkgconfig
make all
```

ndim:
The "proper" way to fix this would be to change the `libgphoto2.pc` and `libgphoto2_port.pc` files to both read

    Cflags: -I${includedir}

instead of

    Cflags: -I${includedir}/gphoto2

msmeissn:
hmm. kamera uses <gphoto2.h>  , needs adaption then.
digikam has also a single <gphoto2.h> include I think.

our tooling (gphoto2, gtkam) uses gphoto2/gphoto2.h already.
python bindings seem to actually work around the above already.
darktable uses gphoto2/gphoto2.h already
wine is ok
... cant check the external tools ...

but i think the change is okay, some packages might need to adjust

ndim:
[...]

PR #764, opened 2022-01-21
pc files: Fix gphoto2 builds, avoid builds with wrong headers

ndim:
In the *.pc files, use a C preprocessor include path both with
and without the gphoto2/ path component:

    Cflags: -I${includedir} -I${includedir}/gphoto2

This allows both the preferred way to include gphoto2 headers

    #include <gphoto2/gphoto2.h>
    #include <gphoto2/gphoto2-camera.h>

and also the early 2000s legacy way (still in use by kamera and
others because we did not properly communicate and enforce that
back when we were still stabilizing the API)

    #include <gphoto2.h>
    #include <gphoto2-camera.h>

to produce builds with the libgphoto2 headers actually corresponding
to the *.pc files in use.

Any other option for the *.pc file Cflags either produces
unintended build failures or builds which use the wrong header
files (the `-I${includedir}/gphoto2` option), or breaks the
`#include <...>` part of the libgphoto2 API from the early 2000s
still in use by kamera and others (the `-I${includedir}` option).

For details, see https://github.com/gphoto/libgphoto2/issues/717 and
https://github.com/gphoto/libgphoto2/issues/717#issuecomment-1018040615

Fixes: https://github.com/gphoto/libgphoto2/issues/717

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,9 @@
 libgphoto2 2.5.28.1 development snapshot
 
+general:
+* fixes build failures of libgphoto2 clients and builds using the
+  wrong libgphoto2 headers (issue #717)
+
 ptp2:
 * fixed a regression in Fuji preview capture which would lead to crashes
 * made the wait-event property extractor more robust (it sometimes gets NULL ptrs), hopefully fixing Sony crashes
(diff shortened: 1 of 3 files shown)


msmeissn (approved):
looks good to me to do it this way... should not disturb othre packages either

PR #776, opened 2022-03-21
BGH1/GH5 : Fixed start / stop recording

barthp12:
`recording start` and `recording stop` were not correctly implemented. 

These commands don't expect data, so `PTP_DP_NODATA` must be used instead of `PTP_DP_SENDDATA`, and the data size must be fixed.


msmeissn:
if you are not sending or receiving DATA, these parameters can be NULL / 0

barthp12:
Indeed, you're right. I don't have a camera today to test, but I'll do that tomorrow and update the pull request

barthp12:
@msmeissn, here the changes. `sendLen` and `**data` are now respectively set to 0 and NULL.

I also squashed the commits.

caseybasichis:
Has anyone tested this with a BS1H?
Trying to get automated control over rec/stop/iso etc.
Wired USB/LAN or wireless is ok.

How can I test this?

msmeissn:
should be in the current libgphoto2 release. 2.5.30.

ks156:
> Has anyone tested this with a BS1H? Trying to get automated control over rec/stop/iso etc. Wired USB/LAN or wireless is ok.
> 
> How can I test this?

Don't know about BSH1, but tested and work with BGH1

Issue #778, opened 2022-03-23
Improve eosremoterelease choices


transistortim:
**Is your feature request related to a problem? Please describe.**
In my setup I need to capture photos with a Canon camera without triggering autofocus, which is why I dug into eosremoterelease.
I found two issues dealing with autofocus and eosremoterelease settings, [here](https://github.com/gphoto/libgphoto2/issues/695) and over at [gphoto2](https://github.com/gphoto/gphoto2/issues/161). After some try & error and problems with localisation I got it working for my use-case.
Nevertheless, in my opinion the eosremoterelease choice names are non-self-explanatory for users not looking into the low-level communication (info on AF is missing, Press 1, 2, 3). I felt invited by the comment [/* debugging */](https://github.com/gphoto/libgphoto2/blob/a06093e4953f1d48bb5fcc937ec7e08242c0fdf5/camlibs/ptp2/config.c#L7463) to do some tests and compile information from various places, see below.

**Describe the solution you'd like**
Draft proposal for new eosremoterelease choices:
choice name         | remotereleaseon | remotereleaseoff
--------------------|-----------------|-----------------
Half-Press no AF    | 1, 1            |
Full-Press no AF*   | 2, 1*           |
Instant-Press no AF | 3, 1            |
Half-Press AF       | 1, 0            |
Full-Press AF       | 2, 0            |
Instant-Press AF    | 3, 0            |
Half-Release        |                 | 1
Full-Release        |                 | 2
Instant-Release     |                 | 3

*remotereleaseon `2, 1` is currently missing from eosremoterelease options. If there is a logic behind this, this should be Full-Press without AF.
"Full-Press" means pressing the button "the other half" (1/2 -> 1) after already having performed a "Half-Press" (0 -> 1/2). "Instant-Press" is the instant press without halting halfway (0 -> 1). I'm not a native speaker, maybe someone knows better fitting terms.

**Describe alternatives you've considered**
I'm open to discussion.

**Additional context**
To understand eosremoterelease, first of all I looked at capturing. What [camera_trigger_canon_eos_capture](https://github.com/gphoto/libgphoto2/blob/a06093e4953f1d48bb5fcc937ec7e08242c0fdf5/camlibs/ptp2/library.c#L5804) does is:

* Regular EOS (not M):
    * `ptp_canon_eos_remotereleaseon (params, 1, 0), _("Canon EOS Half-Press failed")`
[...]

PR #779, opened 2022-03-23
Fix german translation of eosremoterelease choices

transistortim:
In this context release is the opposite of press in the sense of
"loslassen" (see translation of other choices).


transistortim:
Note that for the long term, I propose to generally improve choice names (#778).

Issue #85, opened 2016-09-03
ptp_sony_9280() causes Sony DSC-HX100V reset


supernobby:
Hello!
I have a Sony DSC HX100V and tried gphoto2 2.5.10 form Arch Linux to get the images from it via PTP. However, this fails. The camera does seem to make some kind of reset. In the debug log the last thing that happens is "Sending PTP_OC 0x9280 (Unknown PTP_OC) data..." (probably added for issue #16) and then USB hangs. See the attached Sony_DSC-HX100V-logfile.txt for reference:

[Sony_DSC-HX100V-logfile.txt](https://github.com/gphoto/libgphoto2/files/453526/Sony_DSC-HX100V-logfile.txt)

I then got the source and managed rebuild the ptp2.so which having the calls to ptp_sony_9280() and ptp_sony_9281() disabled. Like that, things seem to work fine, I can e. g. list the files on the camera.

I have so far no knowledge, what ptp_sony_9280() really does. But my DSC-HX100V does not seem to like it, even while claiming, that it supports it. So, I wonder, if something can be done about it.

Andreas

msmeissn:
hmm, thats not so good :(

abrodkin:
So it looks like Sony NEX-3n is affected by similar issue. Is there a need for more logs or any other actions from my side?

mleise:
Sony DSC-HX9V seems to have issues with the command as well: [Sony_DSC-HX9V.txt](https://github.com/gphoto/libgphoto2/files/900965/Sony_DSC-HX9V.txt)
(using libgphoto2-2.5.12)
Unlike the OP nothing hangs. The camera just resets and becomes available again via MTP/PTP.

msmeissn:
I blank out the HX and NEX from sending this opcode combination now

PR #782, opened 2022-03-26
Add Sony Alpha 7 IV (ILCE-7M4) Support

nakato:
This is another case of ptp_sony_9280 causing USB to hang.  Ref #85
Add ILCE-7M4 to the exclude of the ptp_sony_9280 call.

This is followed by two further patches.
One, adding the USB ID for the device, which isn't strictly necessary, but nice to have.
And the other filling in the obvious missing values in CompressionSetting and AspectRatio.

### DIFF
--- a/camlibs/ptp2/cameras/sony-a7m4.txt
+++ b/camlibs/ptp2/cameras/sony-a7m4.txt
@@ -1,0 +1,735 @@
+Camera summary:
+Manufacturer: Sony Corporation
+Model: ILCE-7M4
+  Version: 1.00
+  Serial Number: 00000000000000001234567891234567
+Vendor Extension ID: 0x11 (1.0)
+Vendor Extension Description: Sony PTP Extensions
+
+Capture Formats: 
+Display Formats: Association/Directory, JPEG, ARW, MPEG, Unknown(b982), Unknown(b110)
+
+Device Capabilities:
+	File Download, No File Deletion, File Upload
+	No Image Capture, No Open Capture, Sony Capture
+
+Storage Devices Summary:
+
+Device Property Summary:
+Compression Setting(0x5004):(readwrite) (type=0x2) Enumeration [16,20,19,18,17,4,3,2,1] value: 20
+White Balance(0x5005):(readwrite) (type=0x4) Enumeration [2,4,32785,32784,6,32769,32770,32771,32772,7,32816,32786,32800,32801,32802] value: Automatic (2)
+F-Number(0x5007):(readwrite) (type=0x4) Range [560 - 3200, step 1] value: f/8 (800)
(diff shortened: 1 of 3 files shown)


SiuyuenChow:
@nakato 
hi nakato,
When you use Sony Alpha 7 IV (ILCE-7M4), can you get data normally?
I recently had trouble getting the file path and name.

jaredmdobson:
@SiuyuenChow I can't get file and folders with it.

nakato:
This patch isn't included in a released version of libgphoto2 yet as far as I'm aware.  To use it you'll need to build a development version and ensure the dev library is used over any packaged copy.

I've built and installed a clean copy of this from git just now to check, and I can list folders.  A caveat however is that if access is first attempted with a non-patched copy you have to either send a reset with `--reset` or re-plug the camera, otherwise the camera won't respond to commands.  Some software may attempt to access the camera as soon as it is attached.

I can list folders on my camera, however, the commands are excruciatingly slow, and I have not looked into why.  When I tried to start a transfer to copy the photos off the camera, I saw it was going to take over 24 hours, so I canceled that and I have only transferred 5 photos off the camera with libgphoto2 so far.

For example, this took 2.25min to complete:
```
% time gphoto2 -l
There is 1 folder in folder '/'.                                               
 - store_00010001
There are 67 folders in folder '/store_00010001'.
 - 2022-02-25
...
 - 2022-06-26
There are 0 folders in folder '/store_00010001/2022-02-25'.
There are 0 folders in folder '/store_00010001/2022-02-26'.
...
There are 0 folders in folder '/store_00010001/2022-06-26'.
gphoto2 -l  0.43s user 0.52s system 0% cpu 2:14.70 total
```

I'm not sure what the cause of the slowness is, and I gave up looking into it when I decided I wanted to actually read the PTP standard before really digging into it and found it prohibitively expensive, though I don't think that was directly related to the slowness and more wanting to gain some background on how it works.

SiuyuenChow:
> commands

@nakato 

On SONY, I ran into a problem where I couldn't get the file path and name when listening for photos.

https://github.com/gphoto/libgphoto2/issues/794

sony PTPObjectInfo :
(PTPObjectInfo) oi = {
StorageID = 65536
ObjectFormat = 14337
ProtectionStatus = 0
ObjectCompressedSize = 316688
ThumbFormat = 0
ThumbCompressedSize = 0
ThumbPixWidth = 0
ThumbPixHeight = 0
ImagePixWidth = 0
ImagePixHeight = 0
ImageBitDepth = 0
ParentObject = 0
AssociationType = 0
AssociationDesc = 0
SequenceNumber = 0
Filename = 0x0000000280ca8170 "DSC00152.JPG"
CaptureDate = 0
ModificationDate = 0
Keywords = 0x0000000000000000
}

PR #797, opened 2022-05-13
Fix parallel builds and use them in gh workflow

ndim:
The old gettext 0.18.3 `po/Makefile.in.in` has problems when used with parallel make (`make -jN` with N>1). Running `make -j$(nproc) -l$(nproc) distcheck` on a 12 core system fails reproducibly.

gettext 0.19.1 has apparently fixed those problems with parallel builds, as running `make -j$(nproc) -l$(nproc) distcheck` on the same 12 core system reproducibly succeeds with gettext 0.19.1's `po/Makefile.in.in` file

When parallel builds work, we can now also make use of them during the CI check github Action workflow, so we change all `make` commands there to `make -j$(nproc) -l$(nproc)`.

This is a PR to trigger and test the CI builds, and to allow discussion about requiring gettext 0.19.1 which is a release not even 8 years old, making it the youngest build tool release required to build libgphoto2.


ndim:
While we do not have proper benchmarks from the Github Actions, we can see that adding "-j2 -l2" to the `make` command lines consisitently decreases times, sometimes significantly:

| step              | -j1 | -j2 | sleep iff isatty(1) |
| ----------------- | --- | --- | ------------------- |
| autoreconf        |  24 |  20 |  18                 |
| configure         |  20 |  16 |  15                 |
| make all          |  61 |  27 |  27                 |
| make check        |   4 |   2 |   1                 |
| make install      |   5 |   5 |   4                 |
| make installcheck | n/a | n/a | n/a                 |
| make distcheck    | 137 |  87 |  83                 |
| TOTAL             | 278 | 180 | 176                 |

The few seconds here and there are not significant, but reducing `make all` to about 45% and reducing `make distcheck` to about 60% of its former runtime is a lot better than nothing, even with an `nproc` value of just 2 on the Github Action runner we are running on.

All in all, this reduces the CI build times from a good 4.5 minutes to about 3 minutes, which is a reduction of the total running time by about a third. That is significant.

ndim:
| step              | -j1 | -j2 | sleep iff isatty(1) | --with-camlibs=everything |
| ----------------- | --- | --- | ------------------- | ------------------------- |
| autoreconf        |  24 |  20 |  18                 |  19                       |
| configure         |  20 |  16 |  15                 |  16                       |
| make all          |  61 |  27 |  27                 |  42                       |
| make check        |   4 |   2 |   1                 |   1                       |
| make install      |   5 |   5 |   4                 |  12                       |
| make installcheck | n/a | n/a | n/a                 | n/a                       |
| make distcheck    | 137 |  87 |  83                 | 118                       |
| TOTAL             | 278 | 180 | 176                 | 235                       |

FWIW, with `-j2` we could even build with `--with-camlibs=everything` and still be faster than we were with `-j1` and the default camlib set.

I am not advocating for or against adding `--with-camlibs=everything` to the CI workflow here.

If anything, I would add `DISTCHECK_CONFIGURE_FLAGS=--with-camlibs=everything` to the `make distcheck` command.

msmeissn:
change looks good to me.

PR #822, opened 2022-08-29
ptp2: Add support for Panasonic cameras in camera_trigger_capture

trilader:
Previously there was no support for Panasonic cameras in `camera_trigger_capture` but only in `camera_pansonic_capture` (which gets called by `camera_capture`).

While `camera_capture` works it's not useful for the application I'm working on, as `camera_capture` also downloads the captured image as part of calling it.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -6371,6 +6371,21 @@ camera_trigger_capture (Camera *camera, GPContext *context)
 		return camera_canon_eos_capture (camera, type, path, context);
 	}
 #endif
+
+	if (	(params->deviceinfo.VendorExtensionID == PTP_VENDOR_PANASONIC) &&
+		ptp_operation_issupported(params, PTP_OC_PANASONIC_InitiateCapture)
+	) {
+		PTPContainer	event;
+
+		// Fetch and discard any pending events.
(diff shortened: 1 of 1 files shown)


camlibs/ptp2/library.c
```diff
@@ -6371,6 +6371,21 @@ camera_trigger_capture (Camera *camera, GPContext *context)
...
+	) {
+		PTPContainer	event;
+
+		// Fetch and discard any pending events.
+		C_PTP_REP (ptp_check_event (params));
+		while (ptp_get_one_event(params, &event))
+			; // Do nothing with the events
+
+		// Trigger the actual capture.
+		return ptp_panasonic_capture(params);
```
msmeissn, line 6386: you need to convert the return code to a GP return code:

return translate_ptp_translate_ptp_result(ptp_panasonic_capture(params));

msmeissn (changes requested):

camlibs/ptp2/library.c
trilader: Good point. I didn't think about that and saw `return camera_trigger_canon_eos_capture (camera, context);` further up in the function. But that doesn't have a `ptp_` prefix and therefore should return an already converted result code.

msmeissn:
thanks!

PR #826, opened 2022-09-04
gh workflow: trying to fix libgd-dev dependency issue

ndim:
On the `ubuntu-latest` workers (which appear to be the same as `ubuntu-20.04`, installing `libgd-dev` fails as follows:

```
The following packages have unmet dependencies:
 libgd-dev : Depends: libgd3 (= 2.2.5-5.2ubuntu2.1) but 2.3.3-5+ubuntu20.04.1+deb.sury.org+1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

This tries to fix or work around that.

PR to make the workflow run.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -13,8 +13,14 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: apt-get update
+      run: sudo apt-get update
+    - name: apt-get 3rd party repo libgd-dev dependency workaround
+      run: sudo apt-get remove nginx libgd3
+    - name: apt-get dist-upgrade
+      run: sudo apt-get dist-upgrade
     - name: apt-get install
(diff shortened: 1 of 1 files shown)


ndim:
Same problem on `ubuntu-22.04`:

```
The following packages have unmet dependencies:
 libgd-dev : Depends: libgd3 (= 2.3.0-2ubuntu2) but 2.3.3-5+ubuntu22.04.1+deb.sury.org+1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

ndim:
https://github.com/actions/runner-images/issues/6153

ndim:
As it turns out, the github Ubuntu runners appear to use a 3rd party repo for current php, and that comes depending on its own libgd3 package.

As we do not use php or nginx, we can rust remove the nginx package which pulled in php and libgd packages from that 3rd party repo. If we then install `libgd-dev`, the packages from Ubuntu proper are pulled in and those do not have broken dependencies.

This appears to be a reoccuring problem on github runners, so keeping this workaround in place looks like the way to go for the future.
    
See also https://github.com/actions/runner-images/issues/6153 and recursively follow the linked issues and comments for more information.

PR #833, opened 2022-09-20
Fix `GP_UNKNOWN_EVENT` leaks in examples

RReverser:
Most of the examples give wrong impression that `GP_UNKNOWN_EVENT` data does not need to be freed / is not heap allocated, however it's not the case - those messages are always `malloc`'d by corresponding drivers.


msmeissn:
thanks!

PR #836, opened 2022-09-24
Remove printf leftover

RReverser:
I noticed this in my logs and... according to Git blame it's been there for over 21 years? 😂


msmeissn:
as I am not using win32 based libgphoto2 i never saw it before ;)

RReverser:
That's fair, I just found it funny that a debug leftover can survive that long 😅

PR #838, opened 2022-09-26
Zero-fill CameraFileInfo

RReverser:
I've noticed in my tests that for many files the CameraFileInfo incorrectly had `info.audio` marked as available (with various flags).

Thanks to valgrind I found that it's because `CameraFileInfo` is zero-filled in some places, but not others, so the `audio.fields` (as well as entire `audio`) was filled with uninitialized data.

I tried to find all usages of `info.file.fields = ...` and `info.preview.fields = ...` and zero-initialize them in the same way as it already is in other places.


msmeissn:
thanks!

Issue #166, opened 2017-05-05
Libtool libltdl is not thread safe


kadler15:
I've been doing some work with libgphoto2 accessed from Node.js through a native add-on. I started seeing intermittent segfaults, with stack traces where multiple libuv threads were always centered around some calls to libltdl, usually during abilities list or port info list loading / detection.

E.g.:

```
Program terminated with signal SIGSEGV, Segmentation fault.
__strcmp_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:29
29	../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S: No such file or directory.
(gdb) bt
#0  __strcmp_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:29
#1  0x00007ffff432d10f in ?? () from /usr/lib/x86_64-linux-gnu/libltdl.so.7
#2  0x00007ffff432d504 in ?? () from /usr/lib/x86_64-linux-gnu/libltdl.so.7
#3  0x00007ffff432f5c6 in ?? () from /usr/lib/x86_64-linux-gnu/libltdl.so.7
#4  0x00007ffff432f89d in lt_dlopenadvise () from /usr/lib/x86_64-linux-gnu/libltdl.so.7
#5  0x00007ffff432f9c0 in lt_dlopenext () from /usr/lib/x86_64-linux-gnu/libltdl.so.7
#6  0x00007ffff4744bf6 in gp_abilities_list_load_dir (list=list@entry=0x1e8d4b0, dir=<optimized out>, context=context@entry=0x1ef7da0) at gphoto2-abilities-list.c:210
#7  0x00007ffff4744ec9 in gp_abilities_list_load (list=0x1e8d4b0, context=0x1ef7da0) at gphoto2-abilities-list.c:308

```

I found a relevant piece of info in the ltdl docs:

[Thread safety in libltdl](https://www.gnu.org/software/libtool/manual/html_node/Thread-Safety-in-libltdl.html#Thread-Safety-in-libltdl)

> Libltdl provides a wrapper around whatever dynamic run-time object loading mechanisms are provided by the host system, many of which are themselves not thread safe. Consequently libltdl cannot itself be consistently thread safe.

> If you wish to use libltdl in a multithreaded environment, then you must mutex lock around libltdl calls, since they may in turn be calling non-thread-safe system calls on some target hosts.

[...]

PR #848, opened 2022-10-08
Wrap unsafe libltdl calls with locking

msmeissn:
https://github.com/gphoto/libgphoto2/issues/166

needs review

### DIFF
--- a/libgphoto2/gphoto2-abilities-list.c
+++ b/libgphoto2/gphoto2-abilities-list.c
@@ -36,6 +36,7 @@
 #include <gphoto2/gphoto2-result.h>
 #include <gphoto2/gphoto2-port-log.h>
 #include <gphoto2/gphoto2-library.h>
+#include <gphoto2/gphoto2-port-mutex.h>
 
 #include "libgphoto2/i18n.h"
 
@@ -212,10 +213,12 @@ gp_abilities_list_load_dir (CameraAbilitiesList *list, const char *dir,
 	if (1) { /* a new block in which we can define a temporary variable */
 		foreach_data_t foreach_data = { NULL, GP_OK };
 		foreach_data.list = flist;
+		pthread_mutex_lock(&gpi_libltdl_mutex);
 		lt_dlinit ();
 		lt_dladdsearchdir (dir);
 		ret = lt_dlforeachfile (dir, foreach_func, &foreach_data);
 		lt_dlexit ();
+		pthread_mutex_unlock(&gpi_libltdl_mutex);
 		if (ret != 0) {
 			gp_list_free (flist);
 			GP_LOG_E ("Internal error looking for camlibs (%d)", ret);
@@ -231,7 +234,9 @@ gp_abilities_list_load_dir (CameraAbilitiesList *list, const char *dir,
 		return ret;
 	}
 	GP_LOG_D ("Found %i camera drivers.", count);
+	pthread_mutex_lock(&gpi_libltdl_mutex);
 	lt_dlinit ();
+	pthread_mutex_unlock(&gpi_libltdl_mutex);
 	p = gp_context_progress_start (context, count,
 		_("Loading camera drivers from '%s'..."), dir);
 	for (i = 0; i < count; i++) {
@@ -240,20 +245,28 @@ gp_abilities_list_load_dir (CameraAbilitiesList *list, const char *dir,
 			gp_list_free (flist);
 			return ret;
 		}
+		pthread_mutex_lock(&gpi_libltdl_mutex);
 		lh = lt_dlopenext (filename);
+		pthread_mutex_unlock(&gpi_libltdl_mutex);
 		if (!lh) {
+			pthread_mutex_lock(&gpi_libltdl_mutex);
 			GP_LOG_D ("Failed to load '%s': %s.", filename,
 				lt_dlerror ());
+			pthread_mutex_unlock(&gpi_libltdl_mutex);
 			continue;
 		}
 
 		/* camera_id */
+		pthread_mutex_lock(&gpi_libltdl_mutex);
 		id = lt_dlsym (lh, "camera_id");
+		pthread_mutex_unlock(&gpi_libltdl_mutex);
 		if (!id) {
+			pthread_mutex_lock(&gpi_libltdl_mutex);
 			GP_LOG_D ("Library '%s' does not seem to "
 				"contain a camera_id function: %s",
 				filename, lt_dlerror ());
 			lt_dlclose (lh);
+			pthread_mutex_unlock(&gpi_libltdl_mutex);
 			continue;
 		}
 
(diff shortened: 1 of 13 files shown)


libgphoto2_port/Makefile.am
```diff
@@ -96,6 +96,7 @@ noinst_DATA = libgphoto2_port-uninstalled.pc
 
 # ship headers
 nobase_include_HEADERS =	\
+	gphoto2/gphoto2-port-mutex.h			\
```
ndim, line 99: Why install an internal header file (`gpi_*`)?

ndim (commented):
I will have to examine this on my development workstation as well, not just on my phone.

libgphoto2_port/libgphoto2_port/Makefile.am
```diff
@@ -62,15 +62,15 @@ libgphoto2_port_la_LDFLAGS      += -version-info @LIBGPHOTO2_PORT_VERSION_INFO@
 libgphoto2_port_la_CPPFLAGS     += $(LTDLINCL)
 libgphoto2_port_la_LIBADD       += $(LIBLTDL)
 
-libgphoto2_port_la_LIBADD       += $(INTLLIBS)
```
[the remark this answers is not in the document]
msmeissn: merge issue. removed

libgphoto2/gphoto2-abilities-list.c
```diff
@@ -212,10 +213,12 @@ gp_abilities_list_load_dir (CameraAbilitiesList *list, const char *dir,
 	if (1) { /* a new block in which we can define a temporary variable */
 		foreach_data_t foreach_data = { NULL, GP_OK };
 		foreach_data.list = flist;
+		pthread_mutex_lock(&gpi_libltdl_mutex);
```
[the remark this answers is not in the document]
msmeissn: we are so far not using pthread at all. 
we could put into 2 specific functions, yes

libgphoto2_port/Makefile.am
msmeissn: hmm, we export it for libgphoto2 usage. but probably not needed there

ndim:
@msmeissn Do you mind if I push a few commits resolving the remarks I have made above, and then we can both take a look at the results?

ndim:
I have a pushed a few single commits for easier review, but recommend squashing the commit series into a single commit before committing to master.

ndim:
Two issues remaining:

- [ ] Is the copyright line `Copyright 2017 kadler15 <spurfan15@gmail.com>` good?
- [ ] Is having more than a single `lt_*()' function call in a mutex locked region, required, problematic, or just lazy?

msmeissn:
> @msmeissn Do you mind if I push a few commits resolving the remarks I have made above, and then we can both take a look at the results?

just go ahead, thanks. :) (I am currently sick and otherwise also busy on most days... so expect some delays from me.)

msmeissn:
I think abstracting this into 2 functions and not exposing the pthread mutex over the library is fine. 

one thing, is lt_dlclose / lt_dlopenext refcounting? otherwise we could still load/unload code from other threads.
(e.g. think:
- one thread operating already in a ptp driver driven camera
- second thread does autodetection, dlclose//dlopens around including ptp2 driver
will this interact badly?)

ndim:
I am starting to think we should just seriously look into statically linking all the camlibs into the library. :-/

And do get well soon.

ndim:
> one thing, is lt_dlclose / lt_dlopenext refcounting? otherwise we could still load/unload code from other threads. (e.g. think:
> 
>     * one thread operating already in a ptp driver driven camera
> 
>     * second thread does autodetection, dlclose//dlopens around including ptp2 driver
>       will this interact badly?)

Quoting the libtool info page:

>  -- Function: int lt_dlclose (lt_dlhandle HANDLE)
     Decrement the reference count on the module HANDLE.  If it drops to
     zero and no other module depends on this module, then the module is
     unloaded.  Return 0 on success.

Looks like reference counting to me, and should alleviate the concerns you express above.

ndim:
So, code-wise this looks good to me right now.

The question of the copyright notices remains. The modified files have (according to the git history) about twenty authors. Two or three are mentioned in the copyright notice. Just adding original locking author @kadler15 makes them between a quarter and a half of the authors mentioned in the copyright comment, which looks a bit unfair to all the other authors not mentioned.

So I can either

  * First convert the git history into accurate copyright information (from `Author:`, `Committer:`, and possibly a third person mentioned from the commit message itself), and then add kadler15, and then merge this. This is a lot of work.
  * Just add kadler15 to the modified files, and then merge this, ignoring the other authors. Not that much work, but inconsistent.
  * Just merge this without adding any more copyright notices. Also not much work, also inconsistent.

The latter two options mean kicking the can down the road some more. However, I would want a bit of scripting support to adapt *all* source files' comment in the git repo.

libgphoto2/gphoto2-abilities-list.c
```diff
@@ -212,10 +213,12 @@ gp_abilities_list_load_dir (CameraAbilitiesList *list, const char *dir,
 	if (1) { /* a new block in which we can define a temporary variable */
 		foreach_data_t foreach_data = { NULL, GP_OK };
 		foreach_data.list = flist;
+		gpi_libltdl_lock();
```
RReverser, line 216: Looking at frequent locks/unlocks throughout the function, I wonder if it would've been easier to

1. Rename the unmodified function to `gp_abilities_list_load_dir_impl`.
2. Add a function `gp_abilities_list_load_dir` that locks libltdl, invokes `gp_abilities_list_load_dir_impl`, unlocks libltdl and then returns the return code.

RReverser: Same for other functions that are dedicated to loading libs, and consist for ~90% of libltdl calls anyway.

ndim:
JFTR, regarding the copyright notices: https://github.com/gphoto/libgphoto2/issues/851

libgphoto2/gphoto2-abilities-list.c
msmeissn: @ndim what do you think? seems sensible

libgphoto2_port/libgphoto2_port/Makefile.am
```diff
@@ -62,15 +62,17 @@ libgphoto2_port_la_LDFLAGS      += -version-info @LIBGPHOTO2_PORT_VERSION_INFO@
...
+libgphoto2_port_la_LIBADD       += $(INTLLIBS)
```
Sija, line 65: Unrelated change?

msmeissn:
hmm, some drivers might not be threadsafe, but generally i would hope we are.

RReverser:
> hmm, some drivers might not be threadsafe, but generally i would hope we are.

I found a lot of static mutable variables, as well as MT-unsafe C function calls. I started fixing some of them, but there's too much work so gave up - if it helps, I can send a WIP PR together with Clang analysis if someone wants to take over.

ndim:
@RReverser If you could reply with a few helpful example clang analysis commands to start with (for someone unfamiliar with the tool), that would already go a long way.

RReverser:
By default it provides a lot of noise (false positives), let me filter them out a little more and then I'll share. Most problematic code is indeed in various camera drivers, but then, those are inherent to libgphoto2, so to fix thread safety they would need to be fixed too.

I'll share something in a few days.

libgphoto2/gphoto2-abilities-list.c
RReverser: While I'm working on other parts, let's do this anyway ^

libgphoto2_port/libgphoto2_port/Makefile.am
msmeissn: yes. removed

libgphoto2/gphoto2-abilities-list.c
msmeissn: i did this for this function.

PR #856, opened 2022-10-30
Move serial cache from Canon to port library

RReverser:
I've done this as one of changes in the attempt to ensure re-entrancy, but I figured it's standalone enough to deserve a separate review.

This moves the serial cache (the one that reads a larger chunk to reduce overhead from reading a few bytes at a time) from canon.c to the generic serial port library. The actual cache now lives in the `_GPPortPrivateLibrary` structure instead of static variables, ensuring that it can't be accidentally modified from competing threads. Additionally, by virtue of being in the generic serial port implementation, the cache logic can now benefit all cameras that communicate over the serial port instead of just a single function in Canon driver.

Note that I don't have any ability to actually test cameras over the serial port so would appreciate another pair of eyes.

### DIFF
--- a/camlibs/canon/serial.c
+++ b/camlibs/canon/serial.c
@@ -218,28 +218,10 @@ serial_set_timeout (GPPort *gdev, int to)
 static int
 canon_serial_get_byte (GPPort *gdev)
 {
-	static unsigned char cache[512];
-	static unsigned char *cachep = cache;
-	static unsigned char *cachee = cache;
-	int recv;
-
-	/* if still data in cache, get it */
-	if (cachep < cachee) {
-		return (int) *cachep++;
-	}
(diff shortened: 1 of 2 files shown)


msmeissn:
I will get some serial cameras out of storage to verify it works ;)

msmeissn:
still trying to get my serial setup to work again with he current version. not sure why it does not work currently.

RReverser:
Should I close this at this point?

msmeissn:
sorry for dropping the ball on this,

can you perhaps quickly fixup the merge conflict?

RReverser:
Done.

RReverser:
> Should I close this at this point?

Ping - same question I guess :)

Issue #858, opened 2022-11-03
CI build for Windows


ndim:
At this time, we have automatic CI builds for Linux, but not for OSX, Windows or FreeBSD.

This is to track adding a Windows CI build, probably usign Github Actions inside `.github/workflow/`.

ndim:
I find that Windows/MSYS2 builds on the Github runner infrastructure are **very** slow (by around a factor of 10, see also https://github.com/gphoto/libgphoto2/pull/862#issuecomment-1312659545).

Whether that is inherent to Windows/MSYS2 and therefore cannot be solved at all, or whether there might be other build services available with faster Windows build machines, I do not know yet.

PR #862, opened 2022-11-08
Ci build for windows

ndim:
This tries to add a CI build for Windows

Closes https://github.com/gphoto/libgphoto2/issues/858


.github/workflows/ccpp.yml
```diff
@@ -49,3 +58,89 @@ jobs:
...
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    steps:
+      - name: 'git config'
+        run: git config --global core.autocrlf input
+        shell: bash
+
+      - uses: actions/checkout@v2
```
Sija, line 93:
```suggestion
      - uses: actions/checkout@v3
```

ndim: Thanks, will do that.

.github/workflows/ccpp.yml
```diff
@@ -49,3 +58,93 @@ jobs:
...
+      - name: 'autoreconf'
+        run: autoreconf -vis .
+
+      - name: 'configure'
+        env:
+          CXX: no
+        run: ./configure ${COMMON_CONFIGURE_FLAGS} --prefix=$PWD/__prefix
+
+      - name: 'setup make with nproc'
+        run: NPROC=2; echo "MAKE=make -j$NPROC -l$NPROC" >> $GITHUB_ENV
```
Sija, line 126: maybe:

```suggestion
        run: NPROC=2 && echo "MAKE=make -j$NPROC -l$NPROC" >> $GITHUB_ENV
```

or (I don't think it's needed though)

```suggestion
        run: export NPROC=2 && echo "MAKE=make -j$NPROC -l$NPROC" >> $GITHUB_ENV
```

ndim:
I have a few concerns about Windows CI builds related to performance, overall resource usage, and the impact of longer CI builds on CI build usefulness.

A Linux CI build on ubuntu-latest including "make all" (45 seconds), check, install, installcheck, distcheck takes about 4 minutes, so effectively one build takes around 2 minutes.

On a Windows build, the MSYS2 setup part alone takes about 3 minutes, and "make all" alone takes more than 17 minutes (22 times as slow) for a total build time (without distcheck) of more than 26 minutes.

So this turns CI builds from a 4 minute job on ubuntu-latest to essentially a half hour job on windows-latest, and with two to four (realistically two to three) msys2 matrix `sys` values to build, consumes A LOT of Github runner CPU cycles. For reference, a macos-latest CI build (without distcheck) takes a good 8 minutes, and a CodeQL/Analyze run takes just below 7 minutes.

So we used to wait about 7 minutes until there is a red cross or a green tick mark next to a commit... and Windows CI builds turn that into about 4 times that. As the Windows CI builds take **THAT** long, I wonder whether we want Windows CI builds at all.

However, **IF** we do Windows/MSYS2 CI builds however, I would suggest to only start the Windows builds **after** the Linux (and later macOS) builds have succeeded in order to not take any more resources from the already very strained resource of Windows CPU cycles.

.github/workflows/ccpp.yml
```diff
@@ -110,7 +116,148 @@ jobs:
...
+            autoconf
+            automake
+            gettext-devel
+            libtool
+            make
+            pkgconf
+            msys2-w32api-headers
+            msys2-w32api-runtime
+          pacboy:
+            gcc:p
```
kmilos, line 190: Should be "cc" - this will give you gcc in mingw64/ucrt64 ens, but clang in clang64 env

.github/workflows/ccpp.yml
```diff
@@ -110,7 +116,148 @@ jobs:
...
+          install: >-
+            git
+            autoconf
+            automake
+            gettext-devel
+            libtool
+            make
+            pkgconf
+            msys2-w32api-headers
+            msys2-w32api-runtime
```
kmilos, line 188: If building w/ autotools, better install just env specific `autotools:p` through pacman below, it'll pull in whatever needed.

You don't need msys2-w32api-* packages here, as nothing is being build for the msys2 env...

Here's the list of deps for the [MSYS2 repo package](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-libgphoto2?repo=ucrt64).

So in this section, you basically just need `git`, all the other ones belong to the env specific list in pacboy section below.

ndim: We link against `ws2_32.dll` and `msys2-w32api-runtime` contains a `ws2_32.dll`, so I thought `msys2-w32api-runtime` was the package to install. Could you tell me where my train of thought derails here?

`msys2-w32api-headers` is a leftover from trying to fix the mingw32 build which does not find the `WSAstartup()` function in `ws2_32.dll`. However, it did not help, so I can indeed remove it.

kmilos: `ws2_32.dll` is a Windows system library provided by MS, no need to install it.

Basically, msys2 is a POSIX emulation environment (a fork of Cygwin), while mingw64 etc. are native Windows environments one wants to build for. Apart from a couple of choice msys2 tools, one really doesn't need any more development headers and libraries there, everything must come from the native Windows environments.

ndim:
The problem is not that CI is slow in general. The CI jobs running on Linux are running OK, and the OSX jobs are bearable.

The problem is that especially running CI jobs on Windows is very slow. That makes CI less useful for checking that commits still build.

As an aside, the computation minute argument is not all that relevant here. While GitHub does donate our open source project's minutes, we still want to remain a good citizen here, though, and not waste huge amounts of cycles for little gain.

GitHub explicitly says that 1 Windows minute is worth 2 Linux minutes, and 1 OSX minute is worth 10 Linux minutes. So with about 4 Linux minutes, plus 8 OSX minutes (equal to 80 Linux minutes), 26 Windows minutes without distcheck (equal to 56 Linux minutes), one CI build run costs about 84 Linux minutes right now with linux+osx, and about 140 Linux minutes with windows builds added.

You are arguing for Windows builds not only for CI, but to provide finished binary builds for Windows, which means we need to take a few more steps to comply with the LGPL license. We could accomplish that by creating build artefacts for upload to a well known place (a tarball or zip file) which contains both the binaries **and** the sources, so that anybody downloading the binaries also gets the sources at the same time, and we do not need establish a separate method to provide everybody who has such binaries with the source code.

PR #871, opened 2022-11-18
Only format data log string when a data logger func is registered

watsonmw:
This speeds up large file downloads considerably, as the data log string generation is skipped.


msmeissn:
seems this api set is also not threadsafe with global variables.
but the patch does not make it worse I think, and it looks sensible.
perhapsd @ndim can take a second look too

ndim:
> perhapsd @ndim can take a second look too

Will do

PR #877, opened 2022-12-04
Fix SHUTTER_SPEED_1_6_SEC

petabyt:
Tested on my 1300D


msmeissn:
the canon/ driver is not used for the 1300D. The ptp2 driver is used.

there i see in 

static struct deviceproptableu16 canon_shutterspeed[] = {
..
        { "1.6",        0x0033,0 },
..
so it seems not needed there.

But then, it is likely also incorrect for the older EOS (e.g. the 20D or 350D))

PR #878, opened 2022-12-15
serial: Add check after malloc allocation

likunyur:
Adding pointer allocation check may help the program be more robust.


ndim:
Not sure what this gains us.

There are many more `malloc()` calls with unchecked return values in `camlibs/jd11/serial.c` and `camlibs/jd11/decomp.c`.

Why just check this one `malloc()` call?

msmeissn:
Not sure why the original submitter did this, perhaps as a junior level contribution.

The code definitely has more similar problems, but it is not really worth to fix them as its in outdated driver section, its a serial driver, and the leaks are not that big.

PR #884, opened 2023-01-06
Create canon-eos-discovery.md

lclevy:
allows to document discovered Canon PTP commands using Python syntax or not. Suggested format is Markdown instead of plain text.


msmeissn:
i reversed a lot of the commands, not sure when i have the time to add them. but will try

msmeissn:
(basically the names are in ptp.h already)

lclevy:
Yes, you also sometimes documented the number of request args, dataphase and response parameters, thanks!

PR #897, opened 2023-03-02
Use pthread_t abstract type for thead IDs

kraj:
This is needed to complete https://github.com/gphoto/gphoto2/pull/535


kraj:
not needed closing.

PR #911, opened 2023-05-29
Update library.c

rshnk90:


msmeissn:
there is already an entry with the same name and usb ids:
        {"Sony:Alpha-A6000 (Control)",  0x054c, 0x094e, PTP_CAP},

PR #861, opened 2022-11-07
Add support for Fujifilm X-H2

gergelyerdelyi:
Preview and image capture work in both USB TETHER SHOOTING AUTO & FIXED modes.

PR #922, opened 2023-08-31
Add support for Fujifilm X-T5

killpond:
Duplication of work done in https://github.com/gphoto/libgphoto2/pull/861 but for the X-T5


msmeissn:
usuaully we keep the product ids sorted.
if you would sort it , you willl notice we have an X-T5 entry already down below at line 2681.

also detecetd by testsuite that the name is duplicated

killpond:
Thanks @msmeissn Sorry completely missed that!

PR #935, opened 2023-11-15
Serial fixes

colinleroy:
- Allow changing parity mode after connection is opened
- Fix return value of gp_port_serial_write so that bytes written can be --debug-ged
- Fix inverted logic on parity error checking

The first one is because I will need it to add support for the Quicktake 100 and 150. The two other ones are little bugs :)


msmeissn:
thanks!

PR #955, opened 2024-01-24
camlibs/mars: Fix copy-paste mistake in mars_white_balance()

cher-nov:
cc @kilgota


msmeissn:
any specific reason you switched from r to d?

cher-nov:
> any specific reason you switched from r to d?

For clarity, since the variable `r` is used in conjunction with `g` and `b` later in the code, so using it alone adds an excessive connotation.

msmeissn:
thanks!

Issue #970, opened 2024-03-27
Linking failure with clang: version script assignment of 'LIBGPHOTO2_INTERNAL' to symbol 'gpi_gphoto_port_type_map' failed: symbol not defined


NHOrus:
Attempts to build libgphoto2 with clang fails with error:
```
make[4]: Entering directory '/var/tmp/portage/media-libs/libgphoto2-2.5.30/work/libgphoto2-2.5.30-abi_x86_64.amd64/libgphoto2_port/libgphoto2_port'
/bin/sh ../libtool  --tag=CC   --mode=link clang  -O2 -pipe -march=native -Wall -Wmissing-declarations -Wmissing-prototypes -no-undefined -Wl,--version-script=/var/tmp/portage/media-libs/libgphoto2-2.5.30/work/libgphoto2-2.5.30/libgphoto2_port/libgphoto2_port/libgphoto2_port.ver -version-info 13:0:1 -Wl,-O1 -Wl,--as-needed -Wl,-z,pack-relative-relocs -Wl,--as-needed -o libgphoto2_port.la -rpath /usr/lib64 libgphoto2_port_la-gphoto2-port-info-list.lo libgphoto2_port_la-gphoto2-port-log.lo libgphoto2_port_la-gphoto2-port-version.lo libgphoto2_port_la-gphoto2-port.lo libgphoto2_port_la-gphoto2-port-portability.lo libgphoto2_port_la-gphoto2-port-result.lo -lltdl  -lpthread 
[...]

PR #977, opened 2024-05-08
LIBGPHOTO2_INTERNAL: Symbol gpi_gphoto_port_type_map is present

NHOrus:
in one of two versions of code, but not used anywhere. Noted in NEWS as removed. It breaks compilation in some systems.


eli-schwartz:
Would fix #970

Removed on the 25th via 3cf7a1d04dcffa52907026c4c2d35ee8273f3147 8ddcae4ed2bd052318b0849617e22bf92a98dd17

Issue #950, opened 2024-01-10
`gp_camera_exit` returns `-1: Unspecified error` for Fujifilm X100V


ciaron:
**Describe the bug**

Function `gp_camera_exit` returns `-1: Unspecified error` for Fujifilm X100V

This bug manifested itself in my workflow originally via the `rapid-photo-downloader` tool. There is a bug report here: https://github.com/damonlynch/rapid-photo-downloader/issues/117

I was able to reproduce the original problem using with a C program linking only to libgphoto2 (see below)

**Name the camera**
Camera name as shown by gphoto2 --auto-detect or USB IDs

```
Model                          Port                                            
----------------------------------------------------------
Fuji Fujifilm X100V            usb:001,014 
```

**libgphoto2 and gphoto2 version**
output of: gphoto2 --version

```
gphoto2 2.5.28

Copyright (c) 2000-2021 Marcus Meissner and others

gphoto2 comes with NO WARRANTY, to the extent permitted by law. You may
redistribute copies of gphoto2 under the terms of the GNU General Public
License. For more information about these matters, see the files named COPYING.

This version of gphoto2 is using the following software versions and options:
gphoto2         2.5.28         gcc, popt(m), exif, cdk, aa, jpeg, readline
libgphoto2      2.5.28         standard camlibs (SKIPPING docupen lumix), gcc, no ltdl, EXIF
libgphoto2_port 0.12.0         iolibs: disk ptpip serial usb usbdiskdirect usbscsi, gcc, no ltdl, EXIF, USB, serial without locking

```

**To Reproduce**

Compile the following C program:

```
#include <stdio.h>
#include "gphoto2/gphoto2.h"
#include "gphoto2/gphoto2-version.h"

int test() {
    Camera* camera = NULL;
    CameraWidget* config = NULL;
    int error = GP_OK;
    printf("gp_camera_new\n");
    error = gp_camera_new(&camera);
    if (error) return error;
    printf("gp_camera_init\n");
    error = gp_camera_init(camera, NULL);
    if (error) return error;
    printf("get_camera_get_config\n");
    error = gp_camera_get_config(camera, &config, NULL);
    if (error) return error;
    printf("gp_camera_exit\n");
    error = gp_camera_exit(camera, NULL);
    if (error) return error;
    printf("gp_camera_unref\n");
    error = gp_camera_unref(camera); 
    return error;
}

int main() {
    printf("libgphoto2 version %s\n", *gp_library_version(0));
    int result = test();
[...]

PR #986, opened 2024-07-16
add fixes in case `PTP_DPC_FUJI_PriorityMode` not available (e.g. Fujifilm X100V)

ciaron:
Fix had already been applied for one instance (https://github.com/gphoto/libgphoto2/commit/5f9f6c87189d3223fb26d2db9a6a1a825bda9cf2), I've found all instances and applied the same fix.

I believe this fixes (at least) #950


msmeissn:
thanks!

Issue #987, opened 2024-07-26
camera->functions->wait_for_event fails on Nikon Z9


Noxet:
**Describe the bug**

When calling the function `gp_camera_trigger_capture ` followed by a call to `gp_camera_wait_for_event` in order to download the image, the program sometimes crashes with the following message:

```
*** Contexterror ***              
PTP Device Prop Not Supported
'camera->functions->wait_for_event ( camera, timeout, eventtype, eventdata, context)' failed: -1
```

It happens sporadically and no apparent pattern is seen

**Name the camera**
Nikon Z9

**libgphoto2 and gphoto2 version**
libgphoto 2.5.30

**To Reproduce**
For example, use the example code in the repo: `sample-trigger-capture` which will call the function described above

Nanakbsk:
8

Noxet:
> 8

Very helpful comment :)

msmeissn:
(this is fixed in current GIT afaik)

PR #988, opened 2024-07-26
Removed bitmask from devicepropchanged call that crashed using Nikon Z9

Noxet:
Analysis shows that when the camera failed was when a DevicePropCode > 16bit was being used. In this particular case it is the MovieTVLockSetting in the Z9, with the propcode `0x1D00F`. The event.Param1 variable is a uint32, and the function `ptp_generic_getdevicepropdesc` expects a uint32, so the bitmask is a bit strange here, also causing the program to crash since it will result in a non-existing propcode.

I also updated the prints to show the full 32bit values, instead of only 16 bits.
My test program now runs without issue on the Z9.

For issue #987

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -7441,14 +7441,14 @@ handleregular:
 
 		*eventtype = GP_EVENT_UNKNOWN;
 		/* cached devprop should have been flushed I think... */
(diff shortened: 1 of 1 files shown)


msmeissn:
can you provide logfiles?

properties are 16 bit.

msmeissn:
hmm, checked the Z9 SDK just now ... and you are right actually. Nikon started 32bit propcodes.

PR #998, opened 2024-09-04
fix white-space and indentation inconsistencies in ptp2 driver

axxel:
The code (at least in this part of the project) is effectively formated using tabs with a tab-size of 8 as a means for indentation. Since I last worked on this project almost 10 years ago there have been a few places where spaces were introduced. Some obviously with an indent size of 8 in mind, others with an indent size of 4.

This commit tries its best to make the use of tabs consistent again, at least for indenting lines. The "intra line" indentation (white space after the first non-white-space character) is still a bit of a mess.

Note: this a white-space only change.


msmeissn:
thankls!

PR #1002, opened 2024-09-04
fix white-space and indentation inconsistencies in ...

axxel:
* examples
 * gphoto2
 * libgphoto2
 * libgphoto2_port

The code (at least in this part of the project) is effectively formated using tabs with a tab-size of 8 as a means for indentation.

There have been a few places where spaces were introduced. Some obviously with an indent size of 8 in mind, others with an indent size of 4.

This commit tries its best to make the use of tabs consistent again, at least for indenting lines. The "intra line" indentation (white space after the first non-white-space character) is still a bit of a mess.


msmeissn:
thanks!

PR #1010, opened 2024-09-12
introduce helper function aprintf(fmt, ...)

axxel:
This replaces the verbose and error prone pattern of calling
```
    str = malloc(strlen("some value is 12345678")+1);
    sprintf(str, "some value is %d", val)
```
with
```
    str = aprintf(some value is %d", val);
```

This is similar to the GNU function asprintf() but returns the pointer instead of assigning it to the first parameter. It uses the existing gpi_vsnprintf(fmt, va_list) function.

### DIFF
--- a/libgphoto2_port/gphoto2/gphoto2-port-log.h
+++ b/libgphoto2_port/gphoto2/gphoto2-port-log.h
@@ -218,6 +218,19 @@ __attribute__((__format__(printf,4,5)))
   char*
   gpi_vsnprintf (const char* format, va_list args);
 
+static inline char*
+aprintf(const char *fmt, ...)
+{
+	va_list ap;
+	char* res;
+
+	va_start(ap, fmt);
+	res = gpi_vsnprintf(fmt, ap);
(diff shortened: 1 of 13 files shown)


libgphoto2_port/gphoto2/gphoto2-port-log.h
```diff
@@ -218,6 +218,19 @@ __attribute__((__format__(printf,4,5)))
   char*
   gpi_vsnprintf (const char* format, va_list args);
 
+static inline char*
+aprintf(const char *fmt, ...)
```
hfiguiere, line 222: That would have been nice to add the attributes to mark this as a "format" https://gcc.gnu.org/onlinedocs/gcc-13.3.0/gcc/Common-Function-Attributes.html - then the compiler would be able to give warnings. Without this it introduce more possibilities of error.

axxel: will add that. thanks for the hint.

PR #1016, opened 2024-09-15
fix white-space and indentation inconsistencies (third round)

axxel:
This is basically an amendment to #1015, I just realized that I overlooked some places due to a too restrictive regex.

(I hope this subject is not getting old already...)

This is a whitespace only change (if you include newlines as whitespace).


libgphoto2_port/serial/unix.c
```diff
@@ -524,45 +524,45 @@ gp_port_serial_read (GPPort *dev, char *bytes, int size)
...
 		if (dev->settings.serial.parity != GP_PORT_SERIAL_PARITY_OFF) {
-		    unsigned char ffchar[1];
-		    unsigned char nullchar[1];
-
-		    ffchar[0] = 0xff;
-		    nullchar[0] = 0;
-		    now = read (dev->pl->fd, bytes, 1);
-		    if (now < 0)
-			    return GP_ERROR_IO_READ;
-		    if (!memcmp(bytes,ffchar,1)) {
```
Sija, line 535: There are more changes in this file than just whitespace or am I seeing _sth_?

axxel: Technically, this patch contains non-ws changes in the form of changed line breaks (as did my earlier PRs). But this particular file does indeed only contain ws changes as can be seen if you set the github diff-viewer to ignore ws: https://github.com/gphoto/libgphoto2/pull/1016/files/2f3d334422760dfbfa28035ccdaa66d81400f064?diff=unified&w=1#diff-9cedbc5797364f95753f8ee4613a360ba2a0b47ae5fa851ca3feb6aa5d4ee8fc

Thanks for double checking this "boring" and huge PRs, though. ;)

Issue #81, opened 2016-08-29
Canon EOS700D hangs after some days


obi141:
Hi everybody!
I'm using gphoto2 in combination with a Canon EOS700D on a Raspberry Pi 3. Everything works perfectly for near a week and then suddenly gphoto2 can't reach the camera anymore ("No camera found" error).
I take a picture every 10 minutes and save them on a folder on the Raspberry Pi.
The display of the camera is black and even if I push the menue button on the cam there is no reaction.
Even if I disconnect the power of the cam doesn't helps. In the moment when I take a picture manually everything turns normal but only for a couple of days.
I have this issue on gphoto2.5.9 and even on 2.5.10. Does anybody already have encontered this issue?
I made this test with two EOS700D and I get always the same issue.

Thanks in advice,
obi141

obi141:
Some additional infos: I've found a difference between a earlyer bought EOS700D and the actual which I have in use. The actual lens is a EF-S 18-55 IS STM on the older cam is even a EF-S 18-55 but not STM.
Is it possible that this "new" lens makes the issue?

msmeissn:
The lens should not matter here.

Did you find the problem?

obi141:
Hi Marcus,
you're right, I've checked another system and I've found one where I'm 
using the same lens
but gphoto2 and libgphoto2 with version 2.5.8 and there I have no issues.
At the moment I made a downgrade to 2.5.8 on my test system and now I'm 
waiting
if the issue appears again. Unfortunately it happens after some days so 
I have to wait.

I already found one difference between version 2.5.8 and the 2 higher 
versions (2.5.9. + 2.5.10).
On 2.5.8 the display stays off and doesn't turn on if I press for 
example the menu-button on the camera.
It turns only on when I take a picture via gphoto or if I disconnect the 
cam from USB.
Via gphoto everything works well.

On the newer versions ( > version 2.5.8 ) the display stays on and I can 
navigate trought the menu and change settings while the camera is connected.
At the moment when the camera begins to hang the display stays off and 
there is no reaction if i press the menu-button and gphoto2 doesn't find
the camera anymore. If I press the release button or if I disconnect the 
lens for a moment everything turns normal for a couple of days.
The issue persists, no matter if I change options or not.

I closed the issue because of the fact that it must be another problem 
than the described one...

Kind Regards,
Stefan



Am 06.09.2016 um 11:02 schrieb Marcus Meissner:

PR #1020, opened 2024-09-17
ptp2: condense EOS OLCInfo related debug output

axxel:
A full logging excerpt now looks like this
```
event 237: EOS event OLCInfoChanged (size 81, version 0x0b, mask 0x1fff)
    0x00: 49 00 00 00 ff 1f 00 00 01 00 01 01 a0 10 00 a0
    0x10: 01 01 50 20 2c 01 01 00 00 04 01 01 71 00 00 00
    0x20: 00 00 00 01 00 00 00 00 00 00 00 00 00 00 00 00
    0x30: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
    0x40: 00 00 00 00 00 00 00 00 00
event 237: olcmask 0x0001, 2 bytes: 01 00
event 238: olcmask 0x0002, 6 bytes: 01 01 a0 10 00 a0
event 239: olcmask 0x0004, 5 bytes: 01 01 50 20 2c
event 240: olcmask 0x0008, 4 bytes: 01 01 00 00
event 241: olcmask 0x0010, 4 bytes: 04 01 01 71
event 242: olcmask 0x0020, 6 bytes: 00 00 00 00 00 00
event 243: olcmask 0x0040, 8 bytes: 01 00 00 00 00 00 00 00
event 244: olcmask 0x0080, 4 bytes: 00 00 00 00
event 245: olcmask 0x0100, 6 bytes: 00 00 00 00 00 00
event 246: olcmask 0x0200, 5 bytes: 00 00 00 00 00
event 247: olcmask 0x0400, 5 bytes: 00 00 00 00 00
event 248: olcmask 0x0800, 9 bytes: 00 00 00 00 00 00 00 00 00
event 249: mask 0x1000 entry truncated (1 bytes), olcsizes table (8 bytes) wrong?
event 249: olcmask 0x1000, 1 bytes: 00
```

This allows for easy grepping of the log file. See comments for gained insight into olcmask 0x0010 entries. The user visible strings info strings of the UNKNOWN events have changed as well, e.g.:
```
UNKNOWN OLCInfo event 0x0010, 4 bytes: 04 02 06 2f
```
It also introduces a convenient ptp_bytes2str() function to render some bytes into a string for immediate printing/copying.

### DIFF
--- a/camlibs/ptp2/ptp-pack.c
+++ b/camlibs/ptp2/ptp-pack.c
@@ -34,6 +34,13 @@
 #include <iconv.h>
 #endif
 
+#ifndef MAX
+# define MAX(a, b) ((a) > (b) ? (a) : (b))
+#endif
+#ifndef MIN
+# define MIN(a, b) ((a) < (b) ? (a) : (b))
+#endif
+
 static inline uint16_t
 htod16p (PTPParams *params, uint16_t var)
 {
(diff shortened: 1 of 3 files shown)


axxel:
This revealed an inconsistency in the olcsizes table for the 5Ds: it says 8 for the last entry but my 5Ds only returns 1 byte. This is contradicting [this comment](https://github.com/gphoto/libgphoto2/blob/a26126ade887cb93ac55f3def6b3ab83fe534877/camlibs/ptp2/ptp-pack.c#L2585). Also unclear what issue #81 has to do with that.

msmeissn:
So I think its a size indicator... so the field has no static size, but dynamic and the first byte is the number of bytes following

axxel:
I don't think so because the R8 returns 8 bytes of zero. Looking at all the information available to me at this point, my assumption is that the statement about the 5Ds returning 8 bytes, contradicting the 750D with the same olcversion and the unit on my desk, is simply wrong. You added this comment [here](https://github.com/gphoto/libgphoto2/commit/b1ca489a65f8df796a55ce6b7d84270f9b638837). Do you remember where this information came from? Maybe I can reproduce it with my 5Ds if I configure it differently.

msmeissn:
i dont quite remember anymore. I keep logfiles usually in camlibs/ptp2/cameras/canon*eos* but specifically dont see thast now

PR #1026, opened 2024-09-23
Mostly cosmetic

axxel:
3 random commits, pushing it now because of the first one (fix for a minor display bug I introduced in the last PR).


camlibs/ptp2/ptp.c
```diff
@@ -4628,27 +4628,15 @@ _ptp_sony_getalldevicepropdesc (PTPParams* params, uint16_t opcode)
 		/* debug output to see what changes */
 		if (i != params->nrofdeviceproperties) {
 			switch (dpd.DataType) {
-			case PTP_DTC_INT8:
 #define CHECK_CHANGED(type) \
```
Sija, line 4631: Could this macro definition be moved outside of the scope?

PR #1034, opened 2024-09-30
More ptp2 event logging improvements

axxel:


msmeissn:
(added commit 8b8c81c20555027dbc642f61736d67810ac92a62 after afl showed new crashes)

PR #1040, opened 2024-10-14
More array abstraction use in ptp2

axxel:
More use of the new array abstraction in the ptp2 camlib and some minor cleanups along the way. The PTPParams struct is now completely converted to use the array macros. Also more preparation steps towards merging `dpd_cache` and `canon_props`.

There is a new `TODO` and a `FIXME` comment related to old/potential bugs.

### DIFF
--- a/libgphoto2_port/gphoto2/gphoto2-port-result.h
+++ b/libgphoto2_port/gphoto2/gphoto2-port-result.h
@@ -40,6 +40,13 @@
  * \brief Out of memory
  */
 #define GP_ERROR_NO_MEMORY		-3
+/* FIXME: GP_ERROR_NO_MEMORY is used to communicate two completely differnt
+ * things, which have nothing to do with each other:
+ *  - the camera went out of memory because the storage space ran out, this
+ *    is totally "normal"
+ *  - a malloc on the host computer failed: this will completely interrupt
+ *    the functionality and likely crash the process soonish
+ */
 /**
  * \brief Error in the camera driver
  */
--- a/camlibs/ptp2/Makefile-files
+++ b/camlibs/ptp2/Makefile-files
@@ -32,6 +32,7 @@ ptp2_la_SOURCES      += %reldir%/olympus-wrap.h
 ptp2_la_SOURCES      += %reldir%/chdk.c
 ptp2_la_SOURCES      += %reldir%/fujiptpip.c
 ptp2_la_SOURCES      += %reldir%/ptpip-private.h
+ptp2_la_SOURCES      += %reldir%/array.h
 
 ptp2_la_CFLAGS        = $(camlib_cflags)
 ptp2_la_CPPFLAGS      = $(camlib_cppflags)
--- a/camlibs/ptp2/array.h
+++ b/camlibs/ptp2/array.h
@@ -1,0 +1,121 @@
+/* array.h
+ *
+ * Copyright (C) 2024 Axel Waggershauser <awagger@web.de>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA  02110-1301  USA
+ */
+
+#ifndef CAMLIBS_PTP2_ARRAY_H
+#define CAMLIBS_PTP2_ARRAY_H
+
+#define ARRAYSIZE(ARRAY) (sizeof(ARRAY) / sizeof(ARRAY[0]))
+
+#define move(dst, src) do { \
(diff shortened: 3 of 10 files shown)


libgphoto2_port/gphoto2/gphoto2-port-result.h
```diff
@@ -40,6 +40,13 @@
  * \brief Out of memory
  */
 #define GP_ERROR_NO_MEMORY		-3
+/* FIXME: GP_ERROR_NO_MEMORY is used to communicate two completely differnt
+ * things, which have nothing to do with each other:
+ *  - the camera went out of memory because the storage space ran out, this
+ *    is totally "normal"
+ *  - a malloc on the host computer failed: this will completely interrupt
+ *    the functionality and likely crash the process soonish
+ */
```
Sija, line 49: Perhaps it could be fixed alongside other changes?

msmeissn:
This causes problems found in fuzzing.

==2477413==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf1d03bf4 at pc 0xf2529aff bp 0xffff3f38 sp 0xffff3f2c
READ of size 4 at 0xf1d03bf4 thread T0
    #0 0xf2529afe in find_child ptp2/library.c:8034
    #1 0xf256de6c in get_info_func ptp2/library.c:9378
    #2 0xf78a9761 in gp_filesystem_get_info /home/marcus/projects/afl/libgphoto2/libgphoto2/gphoto2-filesys.c:1867
    #3 0xf787dc7a in gp_camera_file_get_info /home/marcus/projects/afl/libgphoto2/libgphoto2/gphoto2-camera.c:1594
    #4 0x804d0a7 in recursive_directory /home/marcus/projects/afl/libgphoto2/examples/sample-afl.c:79
    #5 0x804cdc9 in recursive_directory /home/marcus/projects/afl/libgphoto2/examples/sample-afl.c:54
    #6 0x804b26f in main /home/marcus/projects/afl/libgphoto2/examples/sample-afl.c:199
    #7 0xf748cd42 in __libc_start_call_main (/lib/libc.so.6+0x24d42) (BuildId: f7c4dd2ccb33969978ca0f5f4ee1df60da37b43b)
    #8 0xf748ce07 in __libc_start_main@GLIBC_2.0 (/lib/libc.so.6+0x24e07) (BuildId: f7c4dd2ccb33969978ca0f5f4ee1df60da37b43b)
    #9 0x804c687 in _start ../sysdeps/i386/start.S:110

0xf1d03bf4 is located 0 bytes after 500-byte region [0xf1d03a00,0xf1d03bf4)
allocated by thread T0 here:
    #0 0xf7a35967  (/lib/libasan.so.8+0xec967) (BuildId: 73a4cd09d0f4b5e83f4566f3b3246a2460ce6d18)
    #1 0xf251a906 in ptp_object_find_or_insert ptp2/ptp.c:9407
    #2 0xf251a906 in ptp_object_find_or_insert ptp2/ptp.c:9360

SUMMARY: AddressSanitizer: heap-buffer-overflow ptp2/library.c:8034 in find_child
Shadow bytes around the buggy address:
  0xf1d03900: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xf1d03980: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0xf1d03a00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0xf1d03a80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00

axxel:
At a first glance, I assume this is caused by the fact that `ptp_object_want` may change the list that we are just iterating over. When working on this, I stumbled upon that fact, which is also mentioned as e.g. "DANGEROUS" in several places throughout the code base. I came to the conclusion that this needs to change and wanted to look into that next. I have not expected that any of my changes made the situation worse with regards to this issue, but it might very well be the case.

I'll look into this soon. Can you maybe give me a hint on how to reproduce your fuzzing?

msmeissn:
I debugged a bit, and it might have been there before already.

The ptp_object_want call ptp_getobjectinfo and that fails in the fuzzer run, and that causes the object to be removed from the list in ptp_object_want (instead of the assumption of always adding to the list).

I think this removal is risky. perhaps better mark it as "no longer present"  object or so (or one that can reappear).

msmeissn:
Reproducer:

in libgphoto2/libgphoto2_port/vusb/vcamera.h  change:

-#undef FUZZING
+#define FUZZING

unzip crash.zip
CFLAGS="-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -O2 -Wall -g" ./configure --enable-vusb --prefix=/usr --libdir=/usr/lib64
make
make check
sudo make install
sudo rm /usr/lib64/libgphoto2_port/0.12.2/usb* 

examples/.libs/sample-afl crash

[crash.zip](https://github.com/user-attachments/files/17413685/crash.zip)

The object array is initially 8 entries, but the find_child loop shrinks it to 4 and then crashes.

I kind of feel the loop tries to remove TWO entries at once somehow, so the ENTRY.val+ENTRY.len!=PTR end of array check does not trigger and it overreads the list.

axxel:
> I think this removal is risky. perhaps better mark it as "no longer present" object or so (or one that can reappear).

That was exactly my thinking as well. And I thought about splitting `ptp_object_want` in a part that might change the list and a part that does not. I have to get a full picture of all it's uses first, though.

msmeissn:
#define for_each(TYPE, PTR, ARRAY) \
-       for (TYPE PTR = ARRAY.val; PTR != ARRAY.val + ARRAY.len; ++PTR)
+       for (TYPE PTR = ARRAY.val; PTR < ARRAY.val + ARRAY.len; ++PTR)

seems to "workaround" it.

PR #1044, opened 2024-10-20
Fix Sony regression by removing ptp_is_vendor_extension_prop

knro:
A git bisect reveled that ptp_is_vendor_extension_prop introduced in https://github.com/gphoto/libgphoto2/commit/f5e51fd63b8b9d76bc5efbfd1f404e722a3b6e78 caused a regression for Sony cameras causing some of them to fail to capture completely.

### DIFF
--- a/camlibs/ptp2/ptp.c
+++ b/camlibs/ptp2/ptp.c
@@ -4620,9 +4620,6 @@ ptp_generic_getdevicepropdesc (PTPParams *params, uint32_t propcode, PTPDevicePr
 		ptp_free_devicepropdesc (dpd_in_cache);
 	}
 
-	if (!ptp_is_vendor_extension_prop(propcode))
-		goto generic;
-
 	if (	(params->deviceinfo.VendorExtensionID == PTP_VENDOR_CANON) &&
 		ptp_operation_issupported(params, PTP_OC_CANON_EOS_RequestDevicePropValue)
 	) {


msmeissn:
@axxel fyi

msmeissn:
I did a different fix, moved the Sony code before the above check, kept the EOS one below it.

Please verify it works.

axxel:
Thanks for bringing this to my attention and sorry for the inconvencience. @knro can you check if the following patch fixes the issue for you?
```patch
diff --git a/camlibs/ptp2/ptp.c b/camlibs/ptp2/ptp.c
index e96603409..01ba8a4b4 100644
--- a/camlibs/ptp2/ptp.c
+++ b/camlibs/ptp2/ptp.c
@@ -4616,10 +4616,8 @@ ptp_generic_getdevicepropdesc (PTPParams *params, uint32_t propcode, PTPDevicePr
                ptp_free_devicepropdesc (dpd_in_cache);
        }
 
-       if (!ptp_is_vendor_extension_prop(propcode))
-               goto generic;
-
-       if (    (params->deviceinfo.VendorExtensionID == PTP_VENDOR_CANON) &&
+       if (    ptp_is_vendor_extension_prop(propcode) &&
+               (params->deviceinfo.VendorExtensionID == PTP_VENDOR_CANON) &&
                ptp_operation_issupported(params, PTP_OC_CANON_EOS_RequestDevicePropValue)
        ) {
                PTPDevicePropDesc *eos_dpd = ptp_find_eos_devicepropdesc(params, propcode);
```

axxel:
> Unfortunately I no longer have access to the Sony camera so cannot test

That is unfortunate. If your PR positively fixed the issue, then I'm confident that @msmeissn's fix did so as well.

msmeissn:
i tested with a Sony A7 and that had capture working now

PR #1054, opened 2024-11-08
In `gp_file_set_data_and_size` free `file->data` only if allocated

Sija:
I've stumbled upon (what I think is) a bug, where I cannot use `gp_file_set_data_and_size` for newly created instances because of: `pointer being freed was not allocated`


ndim:
Isn't calling`free(NULL)` supposed to be a no-op? On what system is that?

Sija:
You're right, I think the error stems from the fact that libgphoto is trying to release the memory it doesn't own (set in `gp_file_set_data_and_size`)

PR #1059, opened 2024-11-19
ptp2/lib: add a check when deleting files

Avenger-285714:
libgphoto2 currently lacks error handling for failed file deletions, potentially leading to unexpected behavior in user-space applications.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -9027,6 +9027,7 @@ delete_file_func (CameraFilesystem *fs, const char *folder,
 			const char *filename, void *data, GPContext *context)
 {
 	Camera *camera = data;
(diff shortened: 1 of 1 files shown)


msmeissn:
this is a bit tricky. See comment above where we ignore the error.

also printf is not the right debugging ;)

Avenger-285714:
> this is a bit tricky. See comment above where we ignore the error.

Hmm...I'm curious about the rationale behind ignoring the errors.
Without error reporting, applications like some file managers that use libgphoto2 will be blindsided by failed deletions.
> 
> also printf is not the right debugging ;)

Yes, yes.
In fact, for this code change, I've referenced the approach used in libmtp...
image

msmeissn:
was merged with another pr

Issue #1032, opened 2024-09-27
How to cross-compile libgphoto2 (to aarch64 on Ubuntu/Debian)?


axxel:
I'm trying to cross-compile to aarch64. I installed `crossbuild-essential-arm64` and now have a working compiler and toolchain, meaning I can compile a hello-world program and run it happily.

Next step was
```sh
configure --build x86_64-pc-linux-gnu --host aarch64-linux-gnu --enable-mpers=check
    --disable-nls --without-jpeg --with-libxml-2.0=no --with-libexif=no --with-libcurl=no --with-gdlib=no
    --disable-serial --with-camlibs=ptp2
    --disable-shared --enable-static  CFLAGS="-g"
```
trying to make it as easy as possible by removing all unnecessary dependencies. configure fails with the following error:
```sh
configure: error: 
libgphoto2 requires libltdl (a part of libtool)
```
Why is `libltdl` still required if I `--disable-shared`?

I'd appreciate any hints towards either:
 * get configure find the missing libltdl
 * compile libgphoto2/libgphoto_port with statically linked ptp2 camlib and libusb driver, to remove the libltdl dependency

Maybe @ndim or @hfiguiere have some relevant insights?

ndim:
Whether you build libgphoto2 and libgphoto2_port as a static or as a dynamic library is not the point here. They both use libltdl to dybamically load camlibs (camera drivers) or iolibs (port drivers), respectively.

axxel:
Thanks for the clarification. Maybe the real question is: why? What is the use case for having a static build that still loads dlls at runtime?

Is there maybe some "hack" that could be applied to the build system to link those symbols statically into the library?

ndim:
The variables to set for building with your custom libltdl:

```
$ ./configure --help | grep -i ltdl
  LTDLINCL    CFLAGS for compiling with libltdl
  LIBLTDL     LIBS to add for linking against libltdl
$ _
```

PR #1065, opened 2024-12-05
Add Meson build system

tintou:
This is the first iteration to add the Meson build system


hfiguiere:
+1 on principle.

msmeissn:
@ndim fyi

ndim:
@msmeissn I know that meson exists, and that is the extent of my knowledge about it.

hfiguiere:
My take is simple. I take meson over antyhing else for a new project build system. I take meson over CMake to migrate away from autotools.

- meson is well supported on the "major" platforms, the trifecta of Linud, macOS and Windows, and supports BSD
- meson is way less obscure than autotools (I know both)
- with ninja as a build runner, it's way faster, and parallel by default

axxel:
I for one certainly [like the idea to replace autotools](https://github.com/gphoto/libgphoto2/pull/731#issuecomment-2355259985) and if @hfiguiere says '+1' instead of 'HELL NO', then maybe meson has a chance ;). I was very interested in Meson when Yussi released it (there might even be some version generation related code from me floating around in the code base ;)), but it was not fulfilling my requirements at the time, so I chose cmake back then. Bottom line: I'm not familiar with how it works nowerdays.

That said: maybe @tintou can comment on the question of whether or not meson could be of help with my cross-compilation question: #1032?

Also, since starting the build setup from scratch is obviously the right time to raise the subject: can we do away with having `libgphoto2_port` as a separate library? And while at it: from an "embedded" developers perspective, it would be preferable to have the option to build a statically linked library with statically linked port and camera drivers (usb + ptp2).

dfandrich:
I not against moving to meson (although I've yet to be convinced the alleged advantages make it worthwhile) but I *am* against having two build systems at the same time (except for a short period during migration). When I've seen that happen in the past, the two very quickly go out of sync making builds inconsistent and creating subtle bugs. Developers also then need to learn two systems instead of one, reducing the pool of people able to fully contribute. If the powers that be decide on meson, automake should be deleted.

tintou:
> I for one certainly [like the idea to replace autotools](https://github.com/gphoto/libgphoto2/pull/731#issuecomment-2355259985) and if @hfiguiere says '+1' instead of 'HELL NO', then maybe meson has a chance ;). I was very interested in Meson when Yussi released it (there might even be some version generation related code from me still floating around in the code base ;)), but it was not fulfilling my requirements at the time, so I chose cmake back then. Bottom line: I'm not familiar with how it works nowerdays.
> 
> That said: maybe @tintou can comment on the question of whether or not meson could be of help with my cross-compilation question: #1032?

Meson has a very extensive [Cross compilation manual](https://mesonbuild.com/Cross-compilation.html), the main difference with Autotools is that everything has to be explicitly set.

> 
> Also, since starting the build setup from scratch is obviously the right time to raise the subject: can we do away with having `libgphoto2_port` as a separate library? And while at it: from an "embedded" developers perspective, it would be preferable to have the option to build a statically linked library with statically linked port and camera drivers (usb + ptp2).

The port I'm showing here is trying to be as close as a 1:1 port from autotools so I'm not doing any change regarding libgphoto2_port being separate, that can be changed afterwards. Using them as a static library is nevertheless possible with the `-Ddefault_library=static` option at configuration time.

tests/meson.build
```diff
@@ -0,0 +1,82 @@
+test_i18n_exe = executable(
+  'test-i18n',
+  'test-i18n.c',
+  dependencies: libgphoto2_dep,
+)
```
eli-schwartz, line 5: If you add `build_by_default: false` then you can skip building these test programs when just running `meson setup && ninja && ninja install` and not ever running `ninja test`. They will automatically be built as part of `ninja test` though.

axxel:
> The port I'm showing here is trying to be as close as a 1:1 port from autotools so I'm not doing any change regarding libgphoto2_port being separate, that can be changed afterwards.

Now that meson is properly up an running and has seen its first set of minor fixes, I'd like to bring back the subject of doing away with `libgphoto2_port` again. In my opinion, this would be a win for the project (reduce complexity for developers, maintainers and users) and now would be the right time, before the meson build option shows up in the [next relase](#1107) - or at least before it completely replacing autotools (which I assume is the goal?).

axxel:
Am I missing something or are the meson files not (yet) part of the release tar-balls? If so, when are they planed to be included in the official release?

PR #1067, opened 2024-12-18
Fixes for Sony A7R3/4/5

watsonmw:
- Fixes for Sony mode 3 (for A7R4/A7R5):
   - capturemode
   - flashmode
   - iso
   - shutterspeed
   - f-number
   - focusmagnify
   - capture-image-and-download
- Allow the following settings to be settable in mode 2 (for A7R3):
   - focusmode
   - flashmode
   - exposure metering
- Re-enabled mode 3 for A7R5, so that:
   - iso enum is reported
   - focus magnify can be set
   - manual focus can be adjusted

### DIFF
--- a/camlibs/ptp2/ptp-pack.c
+++ b/camlibs/ptp2/ptp-pack.c
@@ -734,11 +734,21 @@ outofmemory:
 #define PTP_dpd_Sony_DefaultValue		6
 	/* PTP_dpd_SonyCurrentValue 		6 + sizeof(DataType) */
 
+static uint16_t sony_mode2_settable_props[] = {
+	PTP_DPC_FocusMode,
+	PTP_DPC_ExposureMeteringMode,
+};
+
+static uint16_t sony_mode2_settable_always_props[] = {
+	PTP_DPC_FlashMode,
+};
+
 static inline int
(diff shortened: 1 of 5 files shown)


camlibs/ptp2/ptp-pack.c
```diff
@@ -751,19 +761,56 @@ ptp_unpack_Sony_DPD (PTPParams *params, const unsigned char* data, PTPDeviceProp
...
-		break;
-	case 2: /* display only */
-		dpd->GetSet = 0;	/* just to be safe */
-		break;
+	if (params->sony_mode_ver==2) {
+		/* Old mode */
+		if (dpd->GetSet & 0x80) {
+			/* This is a control value and can be set */
+			/* 0x81 - button, 0x83 - lock, 0x82 - notch, 0x84 - variable */
+			dpd->GetSet = PTP_DPGS_GetSet;
```
msmeissn, line 769: this is falling through isenabled handling ... should there be an else or goto out or something?

watsonmw: Ah ok, maybe... I guess some  'controls' can be (legitimately) disabled, like if the exposure mode precludes that action. For now, I'll add an else - since this needs to be tested per control.

watsonmw:
@msmeissn I made a change for the above issue, any other concerns or guidance?

PR #1070, opened 2024-12-28
Add cross-build

ndim:
Add containerized linux builds for i386, armhf, and aarch64.

Modelled after the avrdude cross builds.


tintou:
You can now use the meson build system to do the cross-build:

On debian-based systems, you can generate the cross-compilation profile with `meson env2mfile --debarch=aarch64 --cross -o my-cross-compilation-profile` for instance and tweak it if required.

Issue #981, opened 2024-06-22
Sony A7 II+ Cameras


tmandys:
I use libgphoto2 (2.5.31) from python-ghoto2 on RPI3/4/5 running Bullseye and Bookworm.  I use more Sony fullframe mirrorless cameras **Sony A7I, II, III, IV**. My application uses **USB preview** or grabbed **HDMI** input to get live preview.  HDMI works rather reliably but USB preview is pain on all Sony models supporting it i.e. **Sony A7 II +**.

### Bug

Without USB preview I can take series of pictures but when USB preview is started then I could take one picture and next one freezes/reboots camera (sequence preview-capture-preview-capture(=shutter)-frozen. Also preview alone stopped working in 30secs or so. 

I investigated and found that USB channel corrupted and caused preview crash (getting typically error code "-1 Bad parameter"). I made fix when automatically reinitializing camera (in 10 cycle loop with n*200ms delay) and opening new session, I see usb address always increments. This algorithm improved preview stability (but still I can see sometimes unresurectable state).

But when I combine with capture and download, then cca every other capture ends with "-7 I/O error" which can be reinitilized often to -1 or minor to OK.  Sometimes the camera is frozen that I must remove battery to reboot, looks like is frozen during shutdown (display is already off).

Tested with **Sony Edge @ MAC, W11** and it works well. Note: The same USB cable as it is often a pain as well

I made Wireshark PTP dumps but no progress yet.

### Camera

- Sony A7 II+
- Note: Canon EOS 6D Mark II works well (it has useless HDMI so USB is the must)

Do you mean is the bug related to PTP protocol or might be issue on USB level ? Note: tested on more RPIs, USB2/3, no difference.

msmeissn:
its likely some form of sony specific protocol issue.

do you have a debuglog?

tmandys:
I found that sequence opensession-preview-capture-preview-capture always fails. As temporary workaround
is possible probably after capture close and open session. The first caoture in preview session
always passes. But seems the reason is waiting to image availability polling _ptp_sony_getalldevicepropdesc vs. event.
gphoto get PTP event too early when data are not availble in camera. Strange I cannot
see event in pcap itself. Is not it related to code below ? Maybe unrelated event or a pending event in queue.

```
#if 1
                /* needed on older cameras like the a58, check for events ... */
                /* This would be unsafe if we get an out-of-order event with no objects present, but
                 * we drained all events above */
[...]

PR #1082, opened 2025-01-31
Sony A7 II&III: Live preview and capture then camera freezes

tmandys:
The bug is at least since 2.5.30 and persists in 2.5.31/head for Sony A7 III (and likely II AFAIR). Issue appears when Live preview is used together with capture image. The first capture passes but next one causes camera freeze/reboot. Bug is in event order evaluation which libgphoto is not familiar with. Note that surprisingly Sony A7 rIV which does not work with 2.5.30 (because of another issue fixed in main branch), does work without this patch in head. So event issue is likely only for III&II models.

Related issue was reported in https://github.com/gphoto/libgphoto2/issues/981

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -4994,9 +4994,13 @@ camera_sony_capture (Camera *camera, CameraCaptureType type, CameraFilePath *pat
 		if (ptp_get_one_event(params, &event)) {
 			GP_LOG_D ("during wait for image event.code=%04x Param1=%08x", event.Code, event.Param1);
 			if (event.Code == PTP_EC_Sony_ObjectAdded) {
-				newobject = event.Param1;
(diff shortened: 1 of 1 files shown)


msmeissn:
i dont really like hardcoding camera names there. is there any other way todetect this

tmandys:
How do you mean to detect it without retesting all Sony models with fix to prove that it won't brake current functionality ? Even I could likely detect somehow what unexpected event causes this issue (price is xxx mandays). But it does not mean that another model will have equal algorithm. I see issue in A7 II & III only. A7 IV does not suffer with it and works with/without bugfix. To make 100% fix I would need reverse engineer/decompile official Sony software if there is not also hardcoded model which is totally beyond my volunteer's budget. I made fix which made gphoto2 usable with A7 II/III to capture image with live preview and we use it for massive digitalization. That's it.

PR #1085, opened 2025-02-02
Some meson fixes

hfiguiere:
libusb detection was flawed, default to libusb1 like the autotools

ltdl need the header, we must check for it.

### DIFF
--- a/libgphoto2_port/meson.build
+++ b/libgphoto2_port/meson.build
@@ -5,29 +5,39 @@ ltdl_dep = cc.find_library('ltdl', required : false)
 if not ltdl_dep.found()
   ltdl_dep = cc.find_library('libltdl', required : false)
 endif
-libusb_dep = dependency('libusb', version: libusb_req, required: 'usb' in get_option('iolibs'))
-libusb1_dep = dependency('libusb-1.0', version: libusb1_req, required: 'libusb1' in get_option('iolibs'))
-lockdev_dep = cc.find_library('lockdev', required : false)
+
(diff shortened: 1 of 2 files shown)


hfiguiere:
cc/ @tintou

libgphoto2_port/meson.build
```diff
@@ -1,34 +1,48 @@
...
+  libusb_dep = dependency('libusb', version: libusb_req, required: 'usb' in get_option('iolibs'))
+  if libusb_dep.found()
+    usb0 = true
+    add_project_arguments('-DHAVE_LIBUSB=1', language: 'c')
+  endif
 endif
 
+lockdev_dep = cc.find_library('lockdev', required : false)
+
 if ltdl_dep.found()
```
tintou, line 46: You set it as required just before 😉

hfiguiere: done.

hfiguiere: Note that it looks like the autotools check make it optional (it's hard to see) but it seems that the include is unconditional. We can revisit this at a later time. Just that I didn't have libltdl instead and couldn't build.

tintou (approved):

tintou:
One question I had: are there systems that do not support libusb1 (i.e. can libusb support be dropped?)

msmeissn:
libusb 1 should meanwhile be everywhere.

I left libusb 0 in out of fear of some other non-standard OS systems like SunOS or so, but i think its time to drop it.

Issue #994, opened 2024-08-26
Please support XDG base directory specification


shmerl:
Currently on Linux libgphoto creates `$HOME/.gphoto` directory when used. Please use XDG base directory specification to avoid cluttering `$HOME`.

shmerl:
Looks like a dup of bug #249, so I'll close this.

shmerl:
Hm, on the other hand that's for the libary, so may be it needs a separate one. So let's keep it open.

ndim:
This is definitely an issue for the libgphoto2 library.

ndim:
So... the libgphoto2 code should, instead of just using `~/.gphoto`

  * Try `$XDG_CONFIG_HOME/gphoto`. If present, use it.
  * Otherwise, look for `~/.gphoto`.
      * If present, make sure the XDG based parent directories exist, creating directories as needed, then rename(2) `~/.gphoto` to `$XDG_CONFIG_HOME/gphoto`. Be aware of symlink attacks and similar things.
      * If not present, create `XDG_CONFIG_HOME/gphoto`. Be aware of symlink attacks and similar things.
  * Use `$XDG_CONFIG_HOME/gphoto`.

This should obviously apply on Linux, BSD.

Should it also apply on MacOS and Windows?

shmerl:
Not exactly. If `$XDG_CONFIG_HOME` is defined - then use `$XDG_CONFIG_HOME/gphoto` (create it if it doesn't exist).

If it's not defined use `$HOME/.config/gphoto` (create it if it doesn't exist).

PR #1108, opened 2025-04-06
Fixing issue 994: respect XDG standards.

eshcheglov:
Use XDG_CONFIG_HOME for storing configuration files

### DIFF
--- a/libgphoto2/gphoto2-setting.c
+++ b/libgphoto2/gphoto2-setting.c
@@ -179,7 +179,10 @@ load_settings (void)
 	SHGetFolderPath(NULL, CSIDL_PROFILE, NULL, 0, buf);
 	strcat (buf, "\\.gphoto");
 #else
-	snprintf (buf, sizeof(buf), "%s/.gphoto", getenv ("HOME"));
+	if (getenv("XDG_CONFIG_HOME") != NULL)
+		snprintf (buf, sizeof(buf), "%s/gphoto", getenv("XDG_CONFIG_HOME"));
+	else
+		snprintf (buf, sizeof(buf), "%s/.config/gphoto", getenv ("HOME"));
 #endif
 	GP_LOG_D ("Creating gphoto config directory ('%s')", buf);
 	(void)gp_system_mkdir (buf);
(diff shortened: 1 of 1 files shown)


eshcheglov:
According to the [XDG specifications](https://specifications.freedesktop.org/basedir-spec/latest/):  

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used. 

Compiled and tested on Ubuntu 25.04.

Fix for [issue 994](https://github.com/gphoto/libgphoto2/issues/994)

hfiguiere:
Existing config is never migrated?

eshcheglov:
> Existing config is never migrated?

Haha, you caught me, I didn't think about that

Okay, I suggest we do this: if the directory `$HOME/.gphoto` already exists, we'll use it. If not - `$XDG_CONFIGS_HOME`. 
This way we won't need to create complicated code for transferring configs, and we'll maintain backward compatibility.

Does that work?

libgphoto2/gphoto2-setting.c
```diff
@@ -179,7 +179,10 @@ load_settings (void)
 	SHGetFolderPath(NULL, CSIDL_PROFILE, NULL, 0, buf);
 	strcat (buf, "\\.gphoto");
 #else
-	snprintf (buf, sizeof(buf), "%s/.gphoto", getenv ("HOME"));
+	if (getenv("XDG_CONFIG_HOME") != NULL)
```
hfiguiere, line 182: Also what if `XDG_CONFIG_HOME` is set but empty? The empty string is returned here.

(same below)

hfiguiere:
> Okay, I suggest we do this: if the directory $HOME/.gphoto already exists, we'll use it. If not - $XDG_CONFIGS_HOME.

No. The new location should be used, and if it doesn't exist, the old one should be **copied** over to the new location, and then used.

eshcheglov:
> No. The new location should be used, and if it doesn't exist, the old one should be copied over to the new location, and then used.

I'm curious why do you think so? 

Copying will lead to user ambiguity because two config files will be present in the system. And this approach also requires config copying during every app start, otherwise, a user who is unaware of the configuration file position change may make changes in `~/` instead of `~/XDG_CONFIG_HOME/`

Moreover, in my opinion, copying approach is less KISS than my proposal

I may refer to the similar issue in [Vim](https://github.com/vim/vim/pull/14182), where they [followed the same](https://github.com/vim/vim/commit/c9df1fb#diff-df93833f5ada39e41ded6f5a6b804b1aa3a75d270848dee769afd9c55f1fcc07R814) path as I suggest

hfiguiere:
> And this approach also requires config copying during every app start

No it doesn't. It's not what I said.

PR #1115, opened 2025-04-28
Allow redirecting `gp_settings_{get,set}` to user defined functions

trilader:
This is useful for applications integrating libgphoto2 which provide their own way of storing user settings.

E.g. in my dayjob I use libgphoto in Qt based applications that use QSettings to store their preferences as storing the settings in the users HOME dir isn't always what we want.
There are cases where one application (and it's settings) are used by multiple user accounts on the same PC and per-user settings would be a problem for PTP/IP pairing as the device GUID is stored in the libgphoto2 settings.

I think merging this would require changing AGE/REVISION/CURRENT, right? I think only AGE needs to be adjusted but I don't quite get how that works

### DIFF
--- a/gphoto2/gphoto2-setting.h
+++ b/gphoto2/gphoto2-setting.h
@@ -28,6 +28,9 @@
 extern "C" {
 #endif /* __cplusplus */
 
+typedef int(*gp_settings_func)(char*,char*,char*,void*);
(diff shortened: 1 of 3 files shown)


msmeissn:
(also bumped minor so version)

PR #1119, opened 2025-05-06
Add color temperature, WB adjust for Olympus OMD

trilader:
I added names/defines for a few of the unnamed settings of an OM-D E-M1 Mark 2.

There is still an issue with the WB adjust though: On the camera the range is ±7 however it gets reported as follows:

In my application: 
```
[…]
[RW] [whitebalanceadjusta] WhiteBalance Adjust A (Radio): Choices: ["65529", "65530", "65531", "65532", "65533", "65534", "65535", "0", "1", "2", "3", "4", "5", "6", "7", ] Value: "2"
[RW] [whitebalanceadjustb] WhiteBalance Adjust B (Radio): Choices: ["65529", "65530", "65531", "65532", "65533", "65534", "65535", "0", "1", "2", "3", "4", "5", "6", "7", ] Value: "65534"
[…]
```

In `gphoto2 --summary` (the one from the OS which doesn't have my changes in it):
```
[…]
Property 0xd161:(readwrite) (type=0x4) Enumeration [65529,65530,65531,65532,65533,65534,65535,0,1,2,3,4,5,6,7] value: 2
Property 0xd162:(readwrite) (type=0x4) Enumeration [65529,65530,65531,65532,65533,65534,65535,0,1,2,3,4,5,6,7] value: 65534
[…]
```

I tried setting the property to `PTP_DTC_UINT16` to also get the negative numbers, but that resulted in
`Type of property 'WhiteBalance Adjust A' expected: 0x0003 got: 0x0004`
in the log and the property still looking the same.

What's the best way to handle this? I assume there is a preferred way for this as I don't think this is the first time a quirk like this happens.

### DIFF
--- a/camlibs/ptp2/config.c
+++ b/camlibs/ptp2/config.c
@@ -6290,6 +6290,7 @@ static struct deviceproptableu16 focusmodes[] = {
 	{ N_("Single-Servo AF"),0x8001, PTP_VENDOR_FUJI },
 	{ N_("Continuous-Servo AF"),0x8002, PTP_VENDOR_FUJI },
 
+	{ N_("Preset MF"),	0x8004, PTP_VENDOR_GP_OLYMPUS_OMD },
 	{ N_("C-AF"),		0x8002, PTP_VENDOR_GP_OLYMPUS_OMD },
 	{ N_("S-AF+MF"),	0x8001, PTP_VENDOR_GP_OLYMPUS_OMD },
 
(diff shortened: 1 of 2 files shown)


msmeissn:
for some reasons it reports them as UINT but they would really be INT16 .

There is no easy way to override this in the get_INT getter/setter, you can however write a own INT16_as_UINT16 get/set functuion if youw ant.

PR #1122, opened 2025-05-09
ci: Use Meson build system instead of autotools

tintou:
Contains also several commits to fix meson build system with macos and linux. Do not squash the commits.


libgphoto2_port/tests/meson.build
```diff
@@ -1,3 +1,7 @@
+gp_port_test_env = [
+    'IOLIBS=@0@'.format(':'.join(iolib_paths)),
```
Sija, line 2:
```suggestion
  'IOLIBS=@0@'.format(':'.join(iolib_paths)),
```

PR #1134, opened 2025-07-06
Meson fixes

heftig:
- **meson: Fix konica localization dir**    
  It got installed into `/usr/libgphoto2/` instead of `/usr/share/libgphoto2/`.

- **meson: Fix project version**    
  It wasn't bumped with the 2.5.32 release.


tintou:
Thank you!

PR #1139, opened 2025-07-15
meson: Add missing version info

heftig:
Necessary for compatibility with the autotools+libtool-built libraries.

Also add the missing export of `gp_filesystem_get_storageinfo`.

---

I've tried to replicate what autotools does. I'm not sure the version script for `libgphoto2` is necessary (AFAICT libtool doesn't add any versions), but the one for `libgphoto2_port` is.

In any case, when comparing the products notice I noticed `gp_filesystem_get_storageinfo` was no longer exported, so I added it to `libgphoto2.sym`.


msmeissn:
thanks!

Issue #1159, opened 2025-08-24
OM-1 Mark II ISO support


coldfire22x:
The OM-1 Mark II supports extension ISOs which exceed the max value of a 16-bit integer, so they've created a new 32-bit opcode for ISO:

```
(d1c0 rw u32): Enumeration [4294967295,80,100,200,250,320,400,500,640,800,1000,1250,1600,2000,2500,3200,4000,5000,6400,8000,10000,12800,16000,20000,25600,32000,40000,51200,64000,80000,102400]
```

joriws:
There is something wrong with the code. I am using Mark 1 of OM-1. Libgphoto 2.5.34 recognizes it as OM-1 Mark 2, and still 0xD1C0 is not mapped to iso, so softwares allowing ISO change do not recognize ISO setting is available on camera. Although I am not sure which code is at wrong as I just git clone main and look at the code. There is OLYMPUS_ISO2 or similar configuration for 0xD1C0 "address" but for some reason I does not work with my camera. And the structure/code is so weird that I am clueless where to look at the fix. I hope with your expertise you could tell why mapping does not work.

Camera id: 33a2:0136

Here is gphoto2 shell summary output for my OM-1 with "--camera 'Olympus OM-1'.
[om1_shell_summary.txt](https://github.com/user-attachments/files/31510912/om1_shell_summary.txt)

PR #1160, opened 2025-08-24
feat: add OM-1 Mark II ISO support

coldfire22x:
Let me know if you'd prefer the Olympus/OM specific additions put somewhere else, thanks!

* add OM-1 Mark II operation code for ISO
* add 32-bit uint compat to *_Olympus_ISO methods

refs: #1159


mszyndel:
I fixed the ISO and a lot of other things in this PR. Can't tell which one's better, feel free to weigh in https://github.com/gphoto/libgphoto2/pull/1292

Issue #1138, opened 2025-07-13
Unable to fetch previews from the camera: Unsupported operation (-6) with Shotwell after upgrading to 2.5.32


aereaux:
**Describe the bug**

After upgrading to libgphoto2 2.5.32 from 2.5.31 I started getting this error when trying to import into shotwell from my camera.

**Name the camera**
```
$ gphoto2 --auto-detect
gphoto2: /usr/lib/libgphoto2_port.so.12: no version information available (required by gphoto2)
Model                          Port                                            
----------------------------------------------------------
Mass Storage Camera            disk:/run/media/aidane/disk
```
This is a Sony A6400, `Bus 001 Device 011: ID 054c:0ca8 Sony Corp. ILCE-6400`

**libgphoto2 and gphoto2 version**
```
$ gphoto2 --version
gphoto2: /usr/lib/libgphoto2_port.so.12: no version information available (required by gphoto2)
gphoto2 2.5.28

Copyright (c) 2000-2021 Marcus Meissner and others

gphoto2 comes with NO WARRANTY, to the extent permitted by law. You may
redistribute copies of gphoto2 under the terms of the GNU General Public
License. For more information about these matters, see the files named COPYING.

This version of gphoto2 is using the following software versions and options:
gphoto2         2.5.28         gcc, popt(m), exif, no cdk, no aa, jpeg, readline
libgphoto2      2.5.32         standard camlibs, gcc, ltdl, EXIF
libgphoto2_port 2.5.32         iolibs: disk ptpip serial libusb1 usb usbdiskdirect usbscsi, gcc, ltdl, EXIF, USB, serial without locking
```

**To Reproduce**
1. Connect camera to computer via USB
2. Mount using Nautilus
3. Open shotwell
4. Click "Mass Storage Camera" in the shotwell side panel
5. Receive a popup with the error message: `Unable to fetch previews from the camera: Unsupported operation (-6)`

msmeissn:
There is a mounted drive with a DCIM directory , likely USB stick or memory card which gets detected first and is tried to be used. you could unmount / remove this card.

aereaux:
Shotwell only seems to show the camera in its list of importable media if I mount it.  Is the way shotwell uses this library now wrong after the update?

wezm:
I'm observing the same behaviour. I too have only ever been able to get Shotwell to import from an SD card after mounting it in Nautilus first. If I just plug my card reader in, with the card plugged into it the drive is created:

[...]

PR #1163, opened 2025-09-25
meson: check more functions that weren't checked.

hfiguiere:
These are checked in autoconf and appear in the source code.

This should help with https://github.com/gphoto/libgphoto2/issues/1138


heftig:
@wezm [reported](https://gitlab.archlinux.org/archlinux/packaging/packages/libgphoto2/-/issues/1#note_326507) this patch works.

PR #1170, opened 2025-10-20
Canon EOS R1

AJaeger74:
my test with images and videos from the R1 passed


AJaeger74:
My /usr/lib/udev/hwdb.d/20-gphoto.hwdb contains this:

```
# Canon EOS R1
usb:v04A9p3313*
GPHOTO2_DRIVER=PTP
ID_GPHOTO2=1
```

PR #1176, opened 2025-11-15
Fix compile errors with gcc 14

bkuhls:


edgar-bonet:
The [compilation error][autobuild] @bkuhls is referring to happened on the [Buildroot][] auto-builders, while compiling for Linux with uClibc for mips64el:

```
jl2005c/jl2005bcd_decompress.c:161:46: error: passing argument 3 of 'jpeg_mem_dest' from incompatible pointer type [-Wincompatible-pointer-types]
  161 |         jpeg_mem_dest (&cinfo, &jpeg_header, &jpeg_header_size);
      |                                              ^~~~~~~~~~~~~~~~~
      |                                              |
      |                                              long unsigned int *
In file included from jl2005c/jpeg_memsrcdest.h:4,
                 from jl2005c/jl2005bcd_decompress.c:39:
…/sysroot/usr/include/jpeglib.h:979:28: note: expected 'size_t *' {aka 'unsigned int *'} but argument is of type 'long unsigned int *'
  979 | EXTERN(void) jpeg_mem_dest JPP((j_compress_ptr cinfo,
      |                            ^~~
```

As shown by this error message, the variable `jpeg_header_size` is being passed (by address) as the last argument of `jpeg_mem_dest()`, thus its type should match the type expected by that function, which here is `size_t` (aka `unsigned int`).

This pull request presumably fixes the issue on Linux/mips64el. However, it breaks the build with GCC 14 on Linux/x86\_64:

```
jl2005c/jl2005bcd_decompress.c: In function 'jl2005bcd_decompress':
jl2005c/jl2005bcd_decompress.c:161:46: error: passing argument 3 of 'jpeg_mem_dest' from incompatible pointer type [-Wincompatible-pointer-types]
  161 |         jpeg_mem_dest (&cinfo, &jpeg_header, &jpeg_header_size);
      |                                              ^~~~~~~~~~~~~~~~~
      |                                              |
      |                                              uint32_t * {aka unsigned int *}
In file included from jl2005c/jpeg_memsrcdest.h:4,
                 from jl2005c/jl2005bcd_decompress.c:39:
/usr/include/jpeglib.h:979:28: note: expected 'size_t *' {aka 'long unsigned int *'} but argument is of type 'uint32_t *' {aka 'unsigned int *'}
```

In order to build on every platform, the type of the variable should be exactly what is written in the prototype of `jpeg_mem_dest()`, and not some platform-dependent synonym. Now, the problem is this type depends on which implementation of `jpeg_mem_dest()` is being used:

* libjpeg 9 expects, as seen here, `size_t`
* libjpeg 8 and jpeg-turbo expect `unsigned long`
* libjpeg <8 does not provide this function. In this case libgphoto2 uses [its own implementation][local], which expects `unsigned long`.

A general solution would probably require conditional compilation, maybe checking both `JPEG_LIB_VERSION` and `LIBJPEG_TURBO_VERSION`.

[autobuild]: https://autobuild.buildroot.net/results/57b/57bb5f02209d8216f0c472d39b15603e0f5db3c5/
[Buildroot]: https://buildroot.org/
[local]: ../blob/v2.5.33/camlibs/jl2005c/jpeg_memsrcdest.h#L13-L14

msmeissn:
can you try my test ptch using configure testing? I currently am not able to test it with jpeg9

edgar-bonet:
@msmeissn wrote:
> can you try my test ptch using configure testing?

I did a couple of build-tests.

**Build-test on Ubuntu 24.04 / x86\_64 + libjpeg-turbo8:**

* It builds, with or without commit a1b7bd094a076f851915f347132ebe0b50b8dd2f.

**Build-test on Buildroot 2026.02 / mips64el + libjpeg 9f:**

* Commit a1b7bd094a076f851915f347132ebe0b50b8dd2f does fix the reported build failure.

Sija:
@msmeissn Your commit ca4441f48117c681cecfb3c883115e0546aed985 broke the CI

msmeissn:
> @msmeissn Your commit [ca4441f](https://github.com/gphoto/libgphoto2/commit/ca4441f48117c681cecfb3c883115e0546aed985) broke the CI

meson build was not working, fixed it quickly

PR #1178, opened 2025-11-17
ptp2: fix memory leak in camera_init

axxel:
ptp_list_folder can be called with a NULL as the last parameter. Otherwise the returned array must be freed by the caller. But here we don't use the list of handles anyway.

Also instead of excluding the code for Sony, only perform it for Canon, which was the reason this was re-introduced.


axxel:
This is related to 21d91dd1e71fac4696a46735772fbdb43f625ccd

PR #1191, opened 2025-11-27
Add Korean translation (po/ko.po)

INI-K:
I added the Korean translation file.


msmeissn:
in general we do translations via translationproject.org

INI-K:
Thank you for the clarification. I will forward the Korean translation to the translationproject.org team for review and inclusion.

PR #1195, opened 2025-12-30
Add support for writing raw JD11 images as Adobe DNG

thenickdude:
This changes the file format for raw files downloaded from JD11-clone cameras from PNM to DNG, so the raw images can now be edited in e.g. Photoshop or RawTherapee. 

In order to support writing DNG, I've added an optional dependency to libtiff for the JD11 camera (only). Since this camera is disabled by default (since it's marked obsolete), I think this won't cause any troubles?

~I've also added a new name for a clone of this camera, "DigiPix AXIS N705", which was how it was sold in the Australia/New Zealand market.~

### DIFF
--- a/camlibs/jd11/jd11.c
+++ b/camlibs/jd11/jd11.c
@@ -113,7 +113,6 @@ static int get_file_func (CameraFilesystem *fs, const char *folder,
 	if(image_no < 0)
 		return image_no;
 
-	gp_file_set_mime_type (file, GP_MIME_PNM);
 	switch (type) {
 	case GP_FILE_TYPE_RAW:
 		result = jd11_get_image_full(camera,file,image_no,1,context);
--- a/camlibs/jd11/serial.c
+++ b/camlibs/jd11/serial.c
@@ -23,7 +23,6 @@
 #include "config.h"
 
 #include <stdio.h>
-#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
@@ -37,6 +36,9 @@
 #include "serial.h"
 #include "decomp.h"
 
+#ifdef HAVE_LIBTIFF
+#include "dng.h"
+#endif
 
 #if 0
 static int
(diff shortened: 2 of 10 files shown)


thenickdude:
I've made libtiff fully optional now, if it isn't found then JD11 still builds fine, and it just writes raw using the old PNM format code instead of DNG.

camlibs/jd11/jd11.c
```diff
@@ -87,6 +87,12 @@ int camera_abilities (CameraAbilitiesList *list)
 	 */
 	strcpy(a.model, "IOMagic:MagicImage 420");
 	gp_abilities_list_append(list, a);
+
+	/* Another rebrand, this one sold in Australia and New Zealand
+	 * https://www.digitalkameramuseum.de/en/cameras/item/digipix-axis-digital-camera
+	 */
+	strcpy(a.model, "DigiPix:AXIS N705");
+	gp_abilities_list_append(list, a);
```
hfiguiere: this should be a separate PR.

camlibs/jd11/serial.c
```diff
@@ -505,36 +519,74 @@ jd11_get_image_full(Camera *camera, CameraFile*file, int nr, int raw, GPContext
 		picture_decomp_v2(imagebufs[1], uncomp[1], 320, 480/2);
 		picture_decomp_v2(imagebufs[2], uncomp[2], 320, 480/2);
 	}
+	free(imagebufs[0]);free(imagebufs[1]);free(imagebufs[2]);free(imagebufs);
```
hfiguiere: one per line please.

camlibs/jd11/serial.c
```diff
@@ -545,10 +597,19 @@ jd11_get_image_full(Camera *camera, CameraFile*file, int nr, int raw, GPContext
...
-	free(imagebufs[0]);free(imagebufs[1]);free(imagebufs[2]);free(imagebufs);
+	
+	gp_file_set_mime_type (file, GP_MIME_PNM);
 	gp_file_append(file, (char*)data, 640*480*3);
 	free(data);
-	return GP_OK;
-}
+	result = GP_OK;
+	
+	finish:
```
hfiguiere:
```suggestion
finish:
```

thenickdude:
Thanks for the review, I've applied those changes now.

PR #1197, opened 2025-12-31
ptp2/lib: Fix unchecked return value in delete_file_func

TomWithKernel:
The ptp_deleteobject() call was not checking its return value, which could cause the function to return success even when the delete operation failed.

Now properly check the return value and return GP_ERROR if the delete operation fails, ensuring errors are properly propagated to the caller.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -9140,8 +9140,7 @@ delete_file_func (CameraFilesystem *fs, const char *folder,
 	CR (find_storage_and_handle_from_path(params, folder, &storage, &handle));
 	handle = find_child(params, filename, storage, handle, NULL);
 
-	/* in some cases we return errors ... just ignore them for now */
-	LOG_ON_PTP_E (ptp_deleteobject(params, handle, 0));
+	C_PTP (ptp_deleteobject(params, handle, 0));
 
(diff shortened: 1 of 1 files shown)


msmeissn:
i ignored the error due to some caameras having issues

Sija:
If so, it would be better to ignore errors only for the problematic cameras instead.

TomWithKernel:
I understand that you chose to ignore the error because some cameras can experience issues.  However, even though the error is ignored and the operation continues, this can cause a situation where the file manager using the PTP protocol reports a file as successfully deleted, even though it was not actually deleted.  This leads to a misleading user experience, as users believe the file is gone when it still remains on the device.

I think it's important to properly return an error if the delete operation fails, to avoid this confusion.

TomWithKernel:
Hi @msmeissn, @Sija,

Just wanted to kindly follow up on this PR.  
Is there anything else I can do to help move it forward, or any further changes you'd like me to make?  
Thank you for reviewing and for your time!

Best regards,  
Haowen Tu

msmeissn:
can you switch it to the standard PTP cehcking macro?

C_PTP ( ... ) does all the magic with return already

TomWithKernel:
I’ve updated the patch to use the standard C_PTP(...) macro as requested and force-pushed the branch.

Issue #1218, opened 2026-03-14
Sony PTP/IP shutdown expected


SPK44:
**Describe the bug**

From what I've gathered, Sony cameras expect the PTP/IP connection to send the command to shutdown the camera at the end of the session. Even when running simple commands with gphoto2, such as `--summary`, the camera won't respond to any connection requests after the first one. The camera itself shows, "Shuts down automatically after save" on the screen after the connection succeeds, no matter which operation has been requested and even without the shutdown command being issued.

Other Sony users on a similar project, https://github.com/falk0069/sony-pm-alt/issues/14 seem to be expecting the shutdown behavior to occur. As mentioned on that issue, this behavior used to be present in gphoto2, which was disabled here f1e3b5ae8f9d8258db9a19861749b2bb6ad57b1e, due to problems it caused with the MTP connection. It was introduced here, 726bd1bda062d2bdc566340e4aada1454bfbd29c for PTP/IP.

**Name the camera**

Sony DSC-RX100M6 ID 054C:0830

**libgphoto2 and gphoto2 version**

```
This version of gphoto2 is using the following software versions and options:
gphoto2         2.5.32         gcc, popt(m), exif, no cdk, aa, jpeg, readline
libgphoto2      2.5.33.1       standard camlibs, gcc, no ltdl, EXIF
libgphoto2_port 0.12.2         iolibs: disk ptpip serial usb1 usbdiskdirect usbscsi, gcc, no ltdl, EXIF, USB, serial without locking
```

This was also tested on 2.5.33

**To Reproduce**

```
gphoto2 --port ptpip:<IP> --summary
gphoto2 --port ptpip:<IP> --summary
```

The first command will run, but the second will fail with the following in the debug log,

```
0.131446 ptp_ptpip_connect           (2): connecting to ptpip:<IP>
0.250238 ptp_ptpip_init_command_request(3): ptpip/init_cmd data: (hexdump of 52 bytes)
0000  34 00 00 00 01 00 00 00-00 00 00 00 00 00 00 00  4...............
0010  ff ff bc 24 11 60 f7 64-53 00 70 00 65 00 65 00  ...$.`.dS.p.e.e.
0020  64 00 79 00 53 00 6e 00-61 00 69 00 6c 00 00 00  d.y.S.n.a.i.l...
0030  00 00 01 00            -                         ....············

0.250262 ptp_ptpip_init_command_request(2): return 52 / len 52
[...]

PR #1220, opened 2026-03-14
Sony PTP/IP connection vendor-specific Initialization and De-initialization

### LINKED ISSUE #1219: Sony PTP/IP initialize connection
**Describe the bug**

When using my camera in PTP/IP mode, trying to list or copy files results in the connection being hung, and a timeout occurring.

It seems that newer Sony cameras expect a vendor-specific connection initialization in PTP/IP mode to confirm the connection before any filesystem operations are allowed. Without the initialization the camera screen will show "Connecting ..." with a loading animation despite a connection being established, and won't respond to filesystem requests.

**Name the camera**

Sony DSC-RX100M6 ID 054C:0830

**libgphoto2 and gphoto2 version**

```
This version of gphoto2 is using the following software versions and options:
gphoto2         2.5.32         gcc, popt(m), exif, no cdk, aa, jpeg, readline
libgphoto2      2.5.33.1       standard camlibs, gcc, no ltdl, EXIF
libgphoto2_port 0.12.2         iolibs: disk ptpip serial usb1 usbdiskdirect usbscsi, gcc, no ltdl, EXIF, USB, serial without locking
```

This was also tested on 2.5.33

**To Reproduce**

Run:
```
gphoto2 -P --port ptpip:<IP> --skip-existing
```

and observe the following in the debug log

```
1.302446 ptp                         (2): ptp_list_folder(storage=0x00010001, handle=0x00040000)
1.302448 ptp                         (2): ptp_list_folder(storage=0x00010001, handle=0x00040005)
1.302451 ptp_ptpip_sendreq           (2): Sending PTP_OC 0x1007 (Get object handles) (0x10001,0x0,0x40005) request...
1.302468 ptp_ptpip_sendreq           (3): ptpip/oprequest data: (hexdump of 30 bytes)
0000  1e 00 00 00 06 00 00 00-00 00 00 00 07 10 12 00  ................
0010  00 00 01 00 01 00 00 00-00 00 05 00 04 00        ..............··

1.302482 ptp_ptpip_getdata           (2): Reading PTP_OC 0x1007 (Get object handles) data...
3.802543 generic_list_func [library.c:8107](0): 'ptp_list_folder (params, storage, parent, &handles)' failed: PTP Timeout (0x02fa)
3.802556 gp_camera_folder_list_files [gphoto2-camera.c:1473](0): 'gp_filesystem_list_files (camera->fs, folder, list, context)' failed: -10
```

I'll be opening a PR shortly to fix this

SPK44:
Resolves #1219, #1218 

I've re-enabled the shutdown command when in PTP/IP mode, which seems to be correct for both my camera and other Sony cameras from observing the reports of others here https://github.com/falk0069/sony-pm-alt/issues/14. I've also added a 9281 call immediately afterwards since the camera wouldn't actually shutdown until it sent back it's acknowledgement. It is possible older cameras did not require the 9281, and might actually be better off without that call since this worked in previous libgphoto2 versions (2.5.9) with those cameras, but I can't be sure.

I've also added a 9280 and 9281 pair to the camera initialization, which was sniffed from packets sent between the camera and Sony Playmemories running on a VM. Since it seems that older cameras worked fine without this in the past, as evidenced by reports in https://github.com/falk0069/sony-pm-alt, I was going to put this behind a specific `device_flag` for my camera only, but it seems that the device flags namespace has run out of bits. I am unsure if this would negatively affect older cameras. Let me know if you have another idea for keeping it specific to the newer Sony cameras.

Finally I rearranged some parameter names in a header and fixed a bug in the `ptp_sony_9280` function which lead to a segfault. The address of an array is just the pointer to the first element, https://eli.thegreenplace.net/2010/01/11/pointers-to-arrays-in-c/, not a double-pointer. It seems that some other functions fall victim to this too, but I didn't want to modify them all in case I am missing something. I was a little skeptical that all these functions were broken in the field,

- ptp_sigma_fp_snap
- ptp_sigma_fp_clearimagedbsingle
- ptp_panasonic_9415
- ptp_panasonic_setdeviceproperty
- ptp_panasonic_manualfocusdrive
- ptp_panasonic_recordmode
- ptp_panasonic_setcapturetarget

## Notes

While I was sniffing the PTP/IP packets from the camera, I deliberately added delay to the TCP stream so I could watch how the camera reacted to various packets. At least for my camera retransmissions of packets started to occur at ~250ms, but did not affect the overall transfer. The camera would timeout the connection and stop responding if ~500ms of delay was added. It seems not very forgiving, but just wanted to note it down in case others are having intermittent issues.

No part of the PR or investigation used LLMs

Thanks for all of your hard work over the years! This is a fantastic project which I am looking forward to using in the future.

### DIFF
--- a/camlibs/ptp2/library.c
+++ b/camlibs/ptp2/library.c
@@ -3285,8 +3285,10 @@ camera_exit (Camera *camera, GPContext *context)
 			/* Sony cameras expect to be issued this shutdown command after any
 			 * operation in PTPIP mode. */
 			if (camera->port->type == GP_PORT_PTPIP) {
-				if (ptp_operation_issupported(params, 0x9280)) {
+				if (ptp_operation_issupported(params, 0x9280)
+					&& ptp_operation_issupported(params, 0x9280)) {
 					C_PTP (ptp_sony_9280(params, 0x4,0,5,0,0,0,0));
+					C_PTP (ptp_sony_9281(params, 0x4));
 				}
 			}
 			break;
@@ -9865,6 +9867,7 @@ camera_init (Camera *camera, GPContext *context)
 		if (camera->port->type == GP_PORT_PTPIP) {
 			if (ptp_operation_issupported(params, 0x9280)) {
 				C_PTP (ptp_sony_9280(params, 0x4,2,2,0,0,1,1));
+				C_PTP (ptp_sony_9281(params, 0x4));
 			}
 		}
 		/* Other setup already done in fixup_cached_deviceinfo */


camlibs/ptp2/library.c
```diff
@@ -3282,12 +3282,15 @@ camera_exit (Camera *camera, GPContext *context)
...
 		case PTP_VENDOR_SONY:
-#if 0
-			/* if we call this, the camera shuts down on close in MTP mode */
-			if (ptp_operation_issupported(params, 0x9280)) {
-				C_PTP (ptp_sony_9280(params, 0x4,0,5,0,0,0,0));
+			/* Sony cameras expect to be issued this shutdown command after any
+			 * operation in PTPIP mode. */
+			if (camera->port->type == GP_PORT_PTPIP) {
+				if (ptp_operation_issupported(params, 0x9280)
+					&& ptp_operation_issupported(params, 0x9280)) {
```
Sija: Duplicate condition. I guess the 2nd one was meant to check for `0x9281` instead.

Sija: Yup, that would be great!

PR #1291, opened 2026-08-06
ptp2: retry Nikon 1 liveview frames that report InvalidStatus

jbroll:
Nikon 1 bodies answer `PTP_OC_NIKON_GetLiveViewImg` with an empty 12-byte data phase and `PTP_RC_NIKON_InvalidStatus` (0xA004) whenever a frame is not ready yet. The transaction completes normally and the bus recovers on its own, but `camera_capture_preview()` treats the code as fatal, so the second `capture_preview()` of a session fails and continuous liveview is impossible — `gphoto2 --capture-movie` exits after a frame or two with `Movie capture error`.

This retries on `InvalidStatus` alongside `DeviceBusy`. The capture paths already treat the two codes the same way for these bodies (the Nikon 1 comments around `ptp_nikon_capture_sdram()` and the wait loop in `camera_trigger_capture()`); this brings the liveview fetch into line.

### Testing

Nikon 1 V3 (`0x04b0:0x060a`, firmware Ver.1.12) over USB, against `v2.5.34`:

- **unpatched:** `capture_preview()` fails on the second call, every time
- **patched:** 1500 consecutive frames at ~58 fps, 640x426, ~0.43 retries per frame, no degradation over the run

`usbmon` confirms the shape of the failing exchange — the data phase is a bare 12-byte container, the response is `0xA004`, and the following transactions proceed normally with no resynchronisation needed:

```
DATA GetLiveViewImg      txn=13  len=12
RESP NIKON_InvalidStatus txn=13
RESP OK                  txn=14
DATA GetLiveViewImg      txn=15  len=13969
RESP OK                  txn=15
```

I don't have a J5 or other Nikon 1 body to hand, so this is verified on the V3 only. The J5 doesn't advertise `0x9203` over USB at all, so it can't reach this path.


jbroll:
Closing — I'll carry this in a fork for now rather than ask upstream to take it.

msmeissn:
not sure why you closed it, it looks sensible. i added the change

jbroll:
My code agent pushed the PR without consulting me and I had not looked at it.  Thank you.

PR #1292, opened 2026-08-06
Describe more OM-1 settings

mszyndel:
I needed ISO for upcoming eclipse and it didn't work. So I ended up hand-decoding all of those other settings.

Confirmed correct on OM-1 body


mszyndel:
@msmeissn vielen dank for merging! Is there any chance of a release including those?

PR #1300, opened 2026-08-17
ptp2: synthesise Sony mode-3 focus magnifier / remote-key props

jeroenlicht:
Sony mode-3 bodies list Focus Magnifier and Remote Key in DeviceInfo, but SDIO_GetAllExtDevicePropInfo does not return descriptors. The existing widgets then fail to bind. This adds those six propcodes to sony_mode3_controls.
focusmagnifier action tested on A7 IV and A7 V
I haven't tested remotekeyup remotekeydown remotekeyleft remotekeyright

Usage (focusmagnifier is MAG,X,Y on the 640×480 live view; X = 0–639, Y = 0–479; center = 320,240):

gphoto2 --get-config focusmagnifier
# e.g. Off,320,240   5.5,63,129   11,63,51
gphoto2 --set-config focusmagnifier=5.5,320,240
gphoto2 --set-config focusmagnifier=Off,320,240
Off,320,240 turns magnification off and resets the center.


jeroenlicht:
Unfortunately, the focusmagnifier action is also added on Sony A7 III, but there it doesn't work. When reading it returns an integer 2 at full view, and doesn't seem to be writable. 
The focusmagnify, focusmagnifyup/down/left/right functions are not there anymore .They used to be there and working on Sony A7 III with release 2_5_31. So maybe we need another solution for the Sony A7 III to get those old actions back.

Issue #1101, opened 2025-03-14
c5a5644035c56312690710493e6f8579d8a28c6c introduced regression in Canon cameras


knro:
**Describe the bug**

Numerous INDI users reported that ISO settings (among others) are not having any effect anymore. This was tested on Canon 600D. After an extensive git bisect, turns out commit c5a5644035c56312690710493e6f8579d8a28c6c is the cause. Not sure why exactly, but after reverting this commit I can verify that ISO settings as set from the INDI GPhoto driver are working fine now.

msmeissn:
can you also try to go to before commit f5e51fd63b8b9d76bc5efbfd1f404e722a3b6e78 which touched this code?

@axxel fyi

knro:
95dc8815fd28f9bafbbe00c790a4efe2a356ad17 works fine.

knro:
c5a5644035c56312690710493e6f8579d8a28c6c sets ISO just fine as well.

axxel:
> [c5a5644](https://github.com/gphoto/libgphoto2/commit/c5a5644035c56312690710493e6f8579d8a28c6c) sets ISO just fine as well.

This is a direct contradiction to your earlier statement:

> After an extensive git bisect, turns out commit https://github.com/gphoto/libgphoto2/commit/c5a5644035c56312690710493e6f8579d8a28c6c is the cause.

Can you explain?

knro:
Sorry that was a copy/paste error, I meant f5e51fd63b8b9d76bc5efbfd1f404e722a3b6e78 is also fine.

knro:
Sorry for late response. Unfortunately, I was debugging this for a user and I wrote a script but I don't have any logs. I already reverted this in my repo and build packages for StellarMate OS. Multiple users confirmed the issue is indeed fixed after this commit was reverted.

axxel:
I'm not questioning your conclusion that reverting this commit fixed your issue but simply reverting this breaks other intended behavior. I would need to get more input (like the mentioned logs) to hopefully find out what is going on to then fix your issue without breaking the other feature.

I don't have a 600D to test this myself. Any chance one of your users might be able to help out getting to the bottom of this?

DomesticFemBoy:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

Domestos2281337:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

mnigbur:
I just ran into the same issue.

[...]

Issue #1153, opened 2025-08-14
2.5.32 breaks (or changes)  Canon EOS  format and ISO functionality


MurrayVK3:
I upgraded to libgphoto 2.5.32 and can no longer set the image format or ISO on a Canon EOS 550D via Ekos and Indi.

The image format choices change with the later version:

With libgphoto 2.5.30 

>gphoto2 --get-config  /main/imgsettings/imageformat
> 		Label: Image Format                                                            
> 		Readonly: 0
> 		Type: RADIO
> 		Current: Small Normal JPEG
> 		Choice: 0 Large Fine JPEG
> 		Choice: 1 Large Normal JPEG
> 		Choice: 2 Medium Fine JPEG
> 		Choice: 3 Medium Normal JPEG
> 		Choice: 4 Small Fine JPEG
> 		Choice: 5 Small Normal JPEG
> 		Choice: 6 RAW + Large Fine JPEG
> 		Choice: 7 RAW
> 		END
> 

With libgphoto 2.5.32

>gphoto2 --get-config  /main/imgsettings/imageformat
> 		Label: Image Format                                                            
> 		Readonly: 0
> 		Type: RADIO
> 		Current: cS
> 		Choice: 0 L
> 		Choice: 1 cL
> 		Choice: 2 M
> 		Choice: 3 cM
> 		Choice: 4 S
> 		Choice: 5 cS
> 		Choice: 6 RAW + L
> 		Choice: 7 RAW
> 		END
> 

The format menu in Ekos gets populated from the above but with version 2.5.32 it no longer changes the setting in the camera.
Also with version 2.5.32 the ISO menu although correctly populated has no effect on the camera setting.

Reverting to libgphoto version 2.5.30 fixes both the above issues.

Testing with gphoto2 I can successfully set both the format and ISO with either versions of libgphoto2 which suggests the issue may be with Indi or Ekos but upgrading/downgrading either or both those makes no difference. The only way to fix it is to revert libgphoto2 back to 2.5.30.  At that point it doesn't make any difference which version of Indi or Ekos I use, only the version of libgphoto2 matters.

**Name the camera**

> gphoto2 --auto-detect 
>         Model                          Port                                            
>         ----------------------------------------------------------
>         Canon EOS 550D                 usb:002,027  
>         

**With format and ISO working:**

> gphoto2 --version
> gphoto2 2.5.28
> 		
> This version of gphoto2 is using the following software versions and options:
[...]

PR #1310, opened 2026-09-20
ptp2: prefer the Canon EOS property path over the generic one

mnigbur:
On Canon EOS bodies the EOS device property list gets appended to deviceinfo.DeviceProps once capture mode has been entered. have_prop() then returns true for properties such as 0xd103 (EOS ISOSpeed) and 0xd120 (EOS ImageFormat).

_get_config(), _set_config() and camera_lookup_by_property() all test that branch before the Canon EOS specific one, so those properties are subsequently read and written with the generic PTP opcodes. A set goes out as 0x1016 SetDevicePropValue, which the camera refuses with 0x200a PTP Device Prop Not Supported, and the error path does a continue, so the EOS branch is never reached as a fallback. The write silently has no effect while --get-config still reports the requested value.

Prefer the EOS branch whenever the property is present in params->canon_props. have_eos_prop() returns 0 unless both the device and the requested vendor are PTP_VENDOR_CANON, so other vendors are unaffected.

Reproduced on a Canon EOS 600D. Note that capture mode has to be entered in the same process first; a bare --set-config iso=... works:

  gphoto2 --set-config capture=on --set-config iso=100 --set-config iso=6400

Before, the ISO sets go out as 0x1016 and are refused with 0x200a, and the camera keeps its previous ISO. After, they go out as 0x9110 EOS_SetDevicePropValueEx and the camera applies them. Verified against the sensor rather than the config readback, by measuring read noise in bias frames. --list-config output is unchanged, and iso, whitebalance, imageformat, shutterspeed round-trips plus --capture-image still work.

Fixes issue #1101  and probably #1153 aswell


msmeissn:
thanksm looks sensible

axxel:
@mnigbur Very nice! Question: did your analysis reveal which exact changeset (presumably proposed by myself) cause the regression you are fixing here?

mnigbur:
@axxel sure, your fix for the canon vendor props in the summary added the duplicates, which get resolved in order so the eos ones were never reached: https://github.com/gphoto/libgphoto2/commit/f5e51fd63b8b9d76bc5efbfd1f404e722a3b6e78

(and this only got active with the parenthesis fix, without the issupported always returns false: https://github.com/gphoto/libgphoto2/commit/c5a5644035c56312690710493e6f8579d8a28c6c)

